PR #2, opened 2019-03-25
Merge updates and more new features

jasondet:
- change get_hist's bins argument back to bins from nbins because it also accommodates auto-binning algorithm strings as well as arrays of bin edges. documentation string updated.
- move function plotting and fit result printing functions back to utils.py since they are not histogram-specific functions
- don't auto-set bounds in fit_hist prior to poissonLL case because minimize() is fine accepting bounds=None but will barf if the length of bounds doesn't match the length of x0 (the function parameters array)
- explicitly name parameter arguments in the functions. makes the code way simpler and adds additional functionality (see e.g. new function get_par_names(func) in utils.py)
- updated print_fit_results() to optionally take a title and add padding carriage returns. Also pulls the parameter names automagically from the function.
- add root-like style file root.mpl and convenience function in utils.py to set style files in pygama (e.g. pgu.set_plot_style('clint') or pgu.set_plot_style('root'))

### DIFF
--- a/pygama/analysis/peak_fitting.py
+++ b/pygama/analysis/peak_fitting.py
@@ -154,7 +154,6 @@ def radford_peak(x, mu, sigma, hstep, htail, tau, bg0, a=1):
     return (1 - htail) * gauss(x, mu, sig, a) + bg_term + step + le_tail
 
 
-# power-law tail plus gaussian https://en.wikipedia.org/wiki/Crystal_Ball_function
 def xtalball(x, mu, sigma, A, beta, m):
     """
     power-law tail plus gaussian https://en.wikipedia.org/wiki/Crystal_Ball_function
--- a/pygama/analysis/peak_fitting.py
+++ b/pygama/analysis/peak_fitting.py
@@ -110,7 +110,7 @@ def gauss2(x, mu, sigma, A=1):
 
 
 #David's peak shape
-def radford_peak(x, mu, sigma, hstep, htail, tau, bg0, a=1)
+def radford_peak(x, mu, sigma, hstep, htail, tau, bg0, a=1):
     #make sure the fractional amplitude parameters stay reasonable...
     if htail < 0 or htail > 1: return np.zeros_like(x)
     if hstep < 0 or hstep > 1: return np.zeros_like(x)
@@ -126,7 +126,7 @@ def radford_peak(x, mu, sigma, hstep, htail, tau, bg0, a=1)
 
 
 # power-law tail plus gaussian https://en.wikipedia.org/wiki/Crystal_Ball_function
-def xtalball(x, mu, sigma, A, beta, m)
+def xtalball(x, mu, sigma, A, beta, m):
     return A * crystalball.pdf(x, beta, m, loc=mu, scale=sigma)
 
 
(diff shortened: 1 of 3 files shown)


pygama/analysis/peak_fitting.py
```diff
@@ -115,75 +114,45 @@ def neg_poisson_log_like(pars, func, hist, bins, **kwargs):
...
-    # compute the step and the low energy tail
-    step = a * hstep * erfc((x - mu) / (sig * np.sqrt(2)))
-    le_tail = h * htail
-    le_tail *= erfc((x - mu) / (sig * np.sqrt(2)) + sig / (tau * np.sqrt(2)))
-    le_tail *= np.exp((x - mu) / tau)
-    le_tail /= (2 * tau * np.exp(-(sig / (np.sqrt(2) * tau))**2))
+    bg_term = bg0  #+ x*bg1
+    if np.any(bg_term < 0): return np.zeros_like(x)
+    step = a * hstep * erfc((x - mu) / (sigma * np.sqrt(2)))
+    le_tail = a * htail * erfc(
```
wisecg: Why did you revert this?  Mine respects the 80 character limit

pygama/analysis/peak_fitting.py
```diff
@@ -115,75 +114,45 @@ def neg_poisson_log_like(pars, func, hist, bins, **kwargs):
...
+            (x - mu) / tau) / (2 * tau * np.exp(-(sigma /
+                                                  (np.sqrt(2) * tau))**2))
+    return (1 - htail) * gauss(x, mu, sigma, a) + bg_term + step + le_tail
 
-    # add up all the peak shape components
-    return (1 - htail) * gauss(x, mu, sig, a) + bg_term + step + le_tail
 
-
-def xtalball(x, *args):
+# power-law tail plus gaussian https://en.wikipedia.org/wiki/Crystal_Ball_function
```
wisecg: Same comment here.  We don't want comments above functions like this, it's not good style

pygama/analysis/peak_fitting.py
```diff
@@ -115,75 +114,45 @@ def neg_poisson_log_like(pars, func, hist, bins, **kwargs):
     """
     Wrapper to give me poisson neg log likelihoods of a histogram
         ln[ f(x)^n / n! exp(-f(x) ] = const + n ln(f(x)) - f(x)
-    FIXME: bin expected mean mu estimated by f(bin_center)*bin_width. Should
+    Note: bin expected mean mu estimated by f(bin_center)*bin_width. Should
     TODO: add option to integrate function over bin
     """
     mu = func(pgh.get_bin_centers(bins), *pars, **kwargs) * pgh.get_bin_widths(bins)
     return np.sum(mu - hist*np.log(mu))
+<<<<<<< HEAD
```
wisecg: Looks like you missed this one

pygama/analysis/peak_fitting.py
```diff
@@ -115,75 +114,45 @@ def neg_poisson_log_like(pars, func, hist, bins, **kwargs):
...
 
-
-def gauss(x, *args):
+#Define a gaussian distribution and corresponding neg log likelihood
```
wisecg: We don't want comments above functions

pygama/analysis/peak_fitting.py
jasondet: not on purpose sorry.

pygama/analysis/peak_fitting.py
jasondet: man what is going on?

jasondet:
I had already found a couple of those, just committed that. I'll take a look at the others. Just me being sloppy

jasondet:
should be ready now?

wisecg:
Did you commit something?  I don't see any more.  But it might be the review mode obscuring changes?

PR #4, opened 2019-03-26
Add stats in hist plots plus a bug fix

jasondet:


wisecg:
`get_formatted_stats` is neat.  I've made myself get used to the new `.format()` method for strings instead of the `%` one.  

https://pyformat.info/

Something to keep in mind for next time.

PR #7, opened 2019-05-24
new requests

matthewstortini:
some of my files aren't opening here. it may be because they exceed the recommended size limit of 50 MB. They are necessary to run a few of the codes though that I sent in the pdf to you (Clint) in slack. I can try to make some edits to make these smaller if it doesn't work for you.


wisecg:
Sadly, this PR has big data files in it that prevent me from even viewing the "files changed" tab.   I'm going to close this one, please make a new one like I mentioned on Slack.

PR #11, opened 2019-05-27
Histogram plotting and fitting updates

jasondet:


jasondet:
Like I said I think ultimately we might not use this, but I had these updates in my code and thought I should at least push them.

PR #18, opened 2019-07-05
added sis3316 decoder

schwarzmario:
add decoder for sis3316.

### DIFF
--- a/experiments/scarf/byHand.py
+++ b/experiments/scarf/byHand.py
@@ -1,0 +1,59 @@
+#!/usr/bin/env python3
+
+#%matplotlib inline
+# %load_ext snakeviz
+import matplotlib.pyplot as plt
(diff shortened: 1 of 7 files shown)


wisecg:
This looks great!  The only comment I have left is that people at UNC/NCSU are developing a 3316 decoder that works with ORCA, and so we might want to put some more references to "llama3316" in the code in various places, to emphasize that this 3316 decoder won't work with an ORCA file.  
I think we don't really need to deal with that until somebody on the ORCA side develops their own 3316 decoder.

This looks really nice, and I'm going to merge.  Thanks for being patient with my review.

Can you send me a message/email and let me know what your next plans are for pygama?  I'd be interested in having you try out some of the calibration stuff (calibration.py) in the MJ60 folder.  Also if you want to start thinking about SiPM signals, that would be interesting too ... just let me know.

PR #20, opened 2019-07-17
MultiPeakFitter code 

matthewstortini:
The following code shall be used in the following way to obtain parameters necessary for the MJ60 paper. The order the codes should be used in for the MultiPeakFitter is:

1. root_hist.py -- This code allows you to input a t2 dataframe, and create a root histogram, which is necessary for the multipeak fitter. It saves the histogram to the metadata data folder, and the other codes in the fitting routine will call to it from there. In this code you must name your histogram, and also the .root file it will be saved to.

2. MakeGuessTemplate.py -- This code requires you to input your .root histogram made in root_hist.py. You must specify the name of your hist and the name of the .root file, and also name your template and its corresponding .root file. This file will be saved to the metadata folder. You can then input guesses, and the code will plot the histogram + the peak functions with the guesses you gave. One should vary these around and observe the plot until they are happy with their guesses.

3. MultiPeakFitter.py -- This code has you input your .root hist, your .root template, and a name for the output file that will save the fit results. This file will also be saved to the metadata folder. If the fit fails your guesses may be lack-luster, but if they look good then try the fit a handful of times. It works by varying the parameters by setting a random seed, so it will not go through the same algorithm every time. This can cause it to fail one time, and then succeed another. This happened to me when fitting the first histogram I worked with. The final results of the fit looked good though once it finally worked.

4. fitresults.py -- This code reads in the fit results, and does some math to print out various useful things. As of now, it simply prints out the calibration parameters. I will continue working on it so that we can cite errors on our calibration parameters, and make plots of our resolution.

------------------

Updates I'll be making to these codes:

1. I will be adding more peaks to MakeGuessTemplate.py for the MultiPeakFitter to fit. Right now the number of peaks it fits is 4, but I will be increasing that number to probably around 9.

2. I will be adding in further analysis to the fitresults.py code, and I plan on having some of the parameters sent to the database. Though, it may be easier to simply keep them in the .root files for the various runs, and use predefined functions to call the parameters you need. There are 10s of parameters from every fit that can be used to make useful plots, so this may be preferred. If you'd like though, the parameters of your choice could be added to a database - perhaps just the calibration parameters. But maybe the errors on the calibration parameters, the resolution function fit parameters / their errors, etc. can just remain in the .root file. 

3. There may be a better way to set up the codes to name the hists/templates/results .root files. Perhaps we can discuss this as a group to get everyone on the same page, and simply adopt a notation to name these files in a consistent manner. The files I have in my metadata folder are from combining runs 280 through 329 and my files are: run_280_329_hist.root, run_280_329_template.root, run_280_329_fit_results.root. The objects in these .root files have the same names, but without the .root extension obviously.

### DIFF
--- a/experiments/mj60/MakeGuessTemplate.py
+++ b/experiments/mj60/MakeGuessTemplate.py
@@ -1,0 +1,115 @@
+import ROOT
+import numpy as np
+import json, os
+
+def main():
+
+    make_template()
+    draw_template()
+
+def make_template():
+
+    with open("runDB.json") as f:
+        runDB = json.load(f)
+    meta_dir = os.path.expandvars(runDB["meta_dir"])
+
+    fout = ROOT.TFile("{}/run_280_329_template.root".format(meta_dir), "RECREATE")
+    fitter = ROOT.GATMultiPeakFitter("run_280_329_template")
+
+    # add regions and peaks
+    x, x[0] = ROOT.vector('double')(1), 583.187
+    fitter.AddRegion(570, 595, x)
+    x, x[0] = ROOT.vector('double')(1), 727.330
+    fitter.AddRegion(718, 737, x)
+    x, x[0] = ROOT.vector('double')(1), 860.557
+    fitter.AddRegion(850, 870, x)
+    x, x[0] = ROOT.vector('double')(1), 2614.511
+    fitter.AddRegion(2580, 2650, x)
+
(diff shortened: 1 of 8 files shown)


wisecg:
Looks good!  Plz paste your writeup above into the MJ60 analysis doc, we can figure out what to do with it from there.

PR #21, opened 2019-07-17
Updated SIS3316 card to function

zachh12:
SIS3316 digitizer classes needed to have model names changed, essentially that was all I did. This produces waveforms when processed (with headers still attached). The length of the waveform is different than GRETINA/SIS3302 so that needs to be determined and the header values can be sorted.

### DIFF
--- a/experiments/ncsu/runDB.json
+++ b/experiments/ncsu/runDB.json
@@ -1,7 +1,7 @@
 {
-  "loc_dir":"$DATADIR/NCSU",
-  "raw_dir":"$DATADIR/NCSU",
-  "tier_dir":"$DATADIR/NCSU/pygama",
+  "loc_dir":".",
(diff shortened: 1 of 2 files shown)


wisecg:
Hey Zach, thanks for contributing!  So you're overwriting Mario's 3316 decoder, which uses a different software (that's why you had to comment out all his stuff).  So please take your code and create a new digitizer function, SIS3316ORCA or something.

zachh12:
I made a separate class for the ORCA 3316. The waveforms look good. Timestamp/crate/card information is not currently being stored properly.

wisecg:
Thanks Zach.  This looks OK to me, I'm going to merge.  Let me know how the decoder does on data from the test stand, and don't hesitate to do another PR if you need to.

PR #26, opened 2019-08-30
Fixed saving of metadata for SIS3316

schwarzmario:
Metadata of the SIS3316 decoder is now saved correctly during production of tier1.
This fixes the error KeyError: 'No object named SIS3316 in the file' during tier2 production


wisecg:
Thanks Mario!  Looks good, I'm going to merge.

PR #27, opened 2019-09-01
More AvsE and C_eff

timmeh08:


timmeh08:
added code, double_gauss, to find an A/E cut for the Th232 source, which contaminates the DEP with Ac228

PR #35, opened 2019-12-20
SIS3316-LLAMA: fixes & changed to LH5 format

schwarzmario:
- Created fixes for the LLAMA SIS3316 file converter in order to make it work again after renamings happened in upstream.
- Created a new config.json / processing.py in the experiments/scarf folder due to these changes
- Changed output format into LH5. This fixed problems arising for waveforms with samples > 3000 and large files. 
Now a ~ 10 GB file with 5000 sample waveforms processes without problems


wisecg:
Thanks Mario!

PR #42, opened 2020-03-06
Dspupgrade

iguinn:
Added get_dataframe to LH5Table to get a pandas dataframe from an LH5Table
Added ls to LH5Store to list all of the groups in an LH5 file


iguinn:
For ls: I have run multiple times into a situation where I've started up a python shell, want to call LH5Store.read_object, but can't remember what the object name is. It's really only a problem for getting the top level group with read_object, because once you have that the LH5 objects make it easy to look at the full hierarchy. It seemed nice to have something in LH5Store for doing that, but if you don't think it fits there, feel free to get rid of it. Would it be possible to make it so that if you don't give read_object a group name it returns the base group of the file or something like that?

For getting dataframes: sounds like someone is going to need to work on this then. I added it just so that I could look at the output from the dsp files more easily, which is (so far) just a simple table. I don't expect to have a lot of time to do this in the near future though.

PR #43, opened 2020-03-06
minor update

jasondet:


jasondet:
FYI pygama/dsp/wfs.bin is a symlink that it looks like Ian accidentally committed.

PR #62, opened 2020-05-03
functional raw_to_dsp.py with json file in pygama/io

timmeh08:


timmeh08:
renamed Ian's apps/raw_to_dsp.py to process_dsp.py.  json file to use with pygama/io/raw_to_dsp.py moved to example folder

PR #74, opened 2020-06-01
bug fix for null VME card and coherent/ncsu updates

zachh12:
Added error handling for digitizers with no waveforms, ORCA saves digitizer data to metadata and when attempting to process this lead to crashes. The rest of the commits are strictly updating NCSU and coherent experiment processing.


zachh12:
For the last commit: For the llamastruck3316 class would be instantiate before the ORCA version of the card and lead to crashing. This is likely due to the two classes sharing a "key" for the card i.e. "SIS3316" or something of that sort. If the instantiation fails this simply lets pygama move on.

wisecg:
Hey Zach -- I'm fine with merging this PR to the master branch, but be aware that we may be switching over to the "dev" branch soon.  We'll need to update the 3316 code again at that point.  Just a heads up.  
Clint

PR #89, opened 2020-08-17
Waveform Browser

iguinn:
Added a WaveformBrowser class. This can draw:
- waveforms
- transformed waveforms defined in the dsp json files
- lines and legend entries for calculated parameters defined in dsp json files
- selected entries using either a list of indices or a mask (similar to how pandas does cuts)
- multiple waveforms at once, with options to scale and shift them by other parameters
To use, construct an instance as described in the documentation, and then call `browser.draw_next()`, `browser.draw_entry(i_entry)`, or loop through them using for `entries in browser:`

@jasondet I also made some changes to lh5.py, specifically defining __len__ for the array-like classes. For Table, I also had to set the size variable when reading, and make sure all of the contained entries are the same size. You might want to take a look at this and make sure the way I did things makes sense.


jasondet:
Nice Ian. The lh5 changes overlap some of mine but I can resolve the conflicts.

PR #105, opened 2020-10-31
added h5py to the deps

mmatteo:
I think this should be listed as a dep


wisecg:
yep, agreed.  merging it

PR #107, opened 2020-11-03
Dev

iguinn:
Re-implemented trap_filter, trap_norm, asymTrapFilter, and pole_zero filters using nopython mode.
- These filters are no longer using SIMD vectorization, but make a single pass over the waveforms, unlike the implementations using numpy commands. This resulted in a small speedup (x2 on my laptop, 10% on cori).
- This leaves significant room for improvement by vectorizing, using C++/eigen

Added a presumming filter that shrinks a waveform by presumming
- Presumming by a factor of 4 resulted in a 2x improvement in speed if all filters used it; if (more realistically) only a few use it, it results in no speedup
- Presumming changes the clock frequency; currently ProcessingChain is designed to used a fixed sampling frequency. To fix this, we would want to attach a sampling frequency to each waveform

Added charge trapping correction to LPGTA dsp. Use a asymmetric trap filter to find t0, and then use a trap filter to perform the integral above the curve starting at t0. Needs testing!

### DIFF
--- a/experiments/lpgta/LPGTA_dsp.json
+++ b/experiments/lpgta/LPGTA_dsp.json
@@ -73,6 +73,28 @@
       "args":["A_10", "trapE", "AoE"],
       "unit": "1/sample",
       "prereqs": ["A_10", "trapE"]
+    },
+    "tp_max": {
+      "function": "argmax",
+      "module": "numpy",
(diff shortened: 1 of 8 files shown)


wisecg:
Looks good, I'm going to merge

PR #109, opened 2020-11-13
Added WaveformBrowser tutorial

iguinn:
Added a jupyter tutorial of the waveform browser, including multiple examples to demonstrate different use-cases and how to use the different options.

The tutorial uses data from the PGT, which means that it won't run on a system other than cori without changing a few variables to point to the right data. It will also expect people to have the $LEGEND_DATA_DIR environment variable set to the same thing as me (and it must be set in the kernel json file!). These factors are not ideal, so if you have a good idea for how to make the tutorial more universal let me know. It seems like maybe we will want to include a small amount of data somewhere in the repo for the purpose of running tutorials.

I added it to a new directory called tutorials; we had discussed adding it to apps, but since there are useful non-tutorial apps like optimizer.py in there, I think it makes more sense as its own thing, but if you disagree feel free to move it. To avoid the problem of committing all the junk and merge conflicts that are produced from running and committing a jupyter notebook, I have added *.ipynb to .git_ignore. This way people can run the tutorial without worrying about committing it; if they want to actually commit changes to an .ipynb, they will want to first clear the outputs, and then do git add -f.

Other changes in this pull request:
- Improved formatting options for legends in the waveform browser
- Added ct_corr (charge trapping correction parameter) and trapEftp (fixed time pickoff energy) to the LPGTA dsp file


iguinn:
Resubmitted to master branch

PR #110, opened 2020-11-17
Oodles of changes

iguinn:
- Apply python dict as a database to DSP files
- Updates to WF browser; moved to dsp directory
- Performance improvements to some processors
- Added charge trapping correction + FTP energy to LPGTA analysis
- Added jupyter tutorial for WF browser; added a git filter to prevent outputs from being added to commits


wisecg:
After extensive Slack discussion, Ian and I are both happy with the changes.

PR #119, opened 2021-01-11
read_object big update

jasondet:
read_object updated to support the following:
- reading into an obj_buffer starting from a non-zero row (obj_buf_start)
- selecting which fields of a struct or table to read (field_mask)
- concatenating data from a list of files

Also includes database usage in dsp_optimize and a few other minor updates.


jasondet:
FYI: multi-file concatenation does not work unless this h5py bug is fixed (in h5py or manually):
https://github.com/h5py/h5py/issues/1792

jasondet:
clint said I can merge it myself

PR #123, opened 2021-01-22
out[i] should be based on wf values prior to i, not subsequent to it.

jasondet:


jasondet:
@iguinn trap_pickoff was computing a pickoff with a different convention than for trap_filter combined with fixed_time_pickoff. This patch changes trap_pickoff to agree with the others. Can you check it and verify that it's not going to cause other problems before it gets merged?

iguinn:
Haha nice, I was just looking at it and came to the same conclusion. That looks right to me!

PR #126, opened 2021-01-30
merge dev changes into master before tagging

wisecg:
I think the current set of changes in `dev` are pretty close to being OK to merge with master.

I'm opening this but not merging it so that @iguinn and @jasondet can offer their takes on any remaining near-term changes that we want to bring in from `dev` before tagging a stable version of `master`.


jasondet:
This PR fixes the error in fc decoding and so I think it should be merged.

wisecg:
We've got to help Sergey!  ;-)  I agree.

PR #128, opened 2021-02-02
Add FlashCam files fields

Kermaidy:
Add decoding of all FlashCam event fields (related to timings).
To be checked before merging:
- runtime definition (or call it 'evttime'?) -> time of event since beginning of run
- timestamp definition -> time of event since epoch (runtime + timeoffset[2])
- dead time computation (do we want to store the calculated deadtime or only store raw information and compute this at upper tiers?)

### DIFF
--- a/experiments/lh5_update/print_times.py
+++ b/experiments/lh5_update/print_times.py
@@ -1,0 +1,91 @@
+#!/usr/bin/env python3
+import os
+import time
+import h5py
+import numpy as np
+import pandas as pd
+from pprint import pprint
+import matplotlib.pyplot as plt
+
+import pygama.analysis.histograms as pgh
+
+def main():
+    """
+    """
+    hf = h5py.File("/lfs/l1/legend/users/kermaidy/pygama_test/char_data/tier1/char_data-V00048B-th_HS2_lat_psa-run0001-210128T143746_tier1.lh5")
+    
+    # General info
+    print(hf.keys())
+    print(hf['raw'].keys())
+    print(hf['stat'].keys())
+    ievt = hf['/raw/ievt']
+
+    # Statustime fields
+    statustime  = hf['/stat/statustime'];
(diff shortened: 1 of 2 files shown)


wisecg:
@Kermaidy Nice PR!  I'd like to go ahead and merge this now and try it out.   I think your suggestions for labels of the time variables are fine.  I personally prefer this convention:

"runtime" = clock time in seconds between run boundaries
"livetime" = "runtime" - "deadtime". 

I think your labeling agrees with my opinion about how it should work.

I'm fine adding all the extra fields, including the deadtime ones.  They are a valid part of the raw data and may be useful later.  If not we can simply not copy all the new columns into e.g. event files later on.

My only question is about how "timestamp" is defined -- is your "epoch" the unix one, i.e. number of seconds since Jan 1 1970, or is "epoch" the time since the start of the current cycle, or current run?

Kermaidy:
Hi @wisecg,

The `timestamp` is the time since 01 Jan 1970 in sec.nsec:
`timestamp = to_master_sec + ts_pps + ts_ticks/(ts_maxticks + 1)`

The `runtime` is similar but without the time offset added:
`runtime = ts_pps + ts_ticks/(ts_maxticks + 1)`

These two parameters, as well as the `deadtime` are calculable from the extra fields I added. However I thought they may be easier to handle from the user (or newcomer) point of view as one does not have to know the details how to retrieve the time of the event. But one could also argue that this is done at later stages.

Thomas Kihm told me he will discuss these FlashCam timing information at the upcoming DAQ call. Can be a good idea for you to attend it in order to have 1st hand information.

wisecg:
@Kermaidy Jason and I agreed that we should just merge this now so that people waiting on FlashCam time information data can use it (Morgan C, David S, etc).  I'll join the next DAQ call, and we'll keep in mind that some of these fields may change as a result.

PR #129, opened 2021-02-05
Baseline and Decay Tail Characterisation Processing Chain

ggmarshall:


iguinn (changes requested):
Hey George,
With the linear_slope_fit algorithm, I think the algorithm used might not be numerically stable. The problem comes in when you take differences of products of numbers that can get really large (like sum_x2*sum_y - sum_x*sum_xy), where the resulting difference can be smaller than the floating point precision of either of those products.
There actually is already implemented in linear_fit.py a linear fitter, although it does not also return the mean and standard deviation like yours does. It's implemented from numerical recipes in C (ch 15, p 664) http://www.grad.hr/nastava/gs/prg/NumericalRecipesinC.pdf. Unless you're sure this algorithm is stable and is just different from that one, can you switch it over to that?
Thanks,
Ian

wisecg:
Hey @ggmarshall, I think you had an example JSON config file for these new processors, can you include that again?  That's very useful when we want to combine with other JSON processor lists.

I'm not sure what to say yet about the other issues Matteo and Ian raised, could you comment on them a little here?

ggmarshall:
Hi @wisecg, I'll include that back in. In terms of the linear_slope algorithm I believe it is stable when comparing it to the results from the linear_fit already included. I am not sure however if it is needed.

mmatteo:
@wisecg, I think you can go ahead and tag the code w/o this PR. We will keep developing processors and won't use the tag for the hades production

mmatteo:
@ggmarshall, @iguinn, @wisecg, after yesterday discussion, I guess the log_check module return value is okay. 

Concerning the linear_slope_fit, I would compare the value returned by the new module against the old ones. If they are compatible, I would use the new module which should reduce the computing time and make the json file more readable

ggmarshall:
Hi, I re-added the processor list.  I also ran a comparison between the 2 processors on 15 thorium files which was around 5.5 million waveforms. The slope gave no differences, the intercept gave small differences (O(0.001)) for some waveforms (3193 differed in total out of 5673212) . I have also included the percentage difference between the 2 which is of order 10^-5 %.
Screenshot 2021-02-12 at 14 57 54
Screenshot 2021-02-12 at 14 58 01

mmatteo:
Thanks @ggmarshall. I would consider the algorithm as validated. @iguinn, do you agree? Can we go ahead and merge this PR?

Kermaidy:
Hi @ggmarshall,
I suggest you also run it on a PGT data (e.g. run0030) where you get not only nice waveform but also discharges, leakage current. That's a better stability test to my opinion. Although can be done later ;-)
Yoann

ggmarshall:
Hi I've switched the intercept finding over to the same method in the linear fit processor so no differences now.

pygama/dsp/_processors/min_max.py
```diff
@@ -0,0 +1,28 @@
...
+
+def min_max(wf, argmin, argmax, wf_min, wf_max, ):
+    '''
+    Finds the min, max and their time position for a waveform
+    '''
+
+    
+    wf_min[0] = wf[0]
+    wf_max[0] = wf[0]
+    argmin[0] = 0
```
mmatteo, line 19: I would suggest to inititialize argmin to `argmin[0]=float(math.inf)` and argmax[0]=float(-math.inf)`

iguinn (approved):

iguinn:
Hi George, thanks for checking this! I'm satisfied.

pygama/dsp/_processors/min_max.py
ggmarshall: done

mmatteo (approved):
Looks good! Thanks for the extra work @ggmarshall

mmatteo:
@wisecg, do you have any other request?  can we merge this PR?

wisecg:
@mmatteo @ggmarshall We're getting merge conflicts here, can one of you take a look so I can merge?

mmatteo:
@wisecg, this is ready to go

PR #135, opened 2021-02-11
Output fcio_config. Also rename stat -> fcio_status

jasondet:


Kermaidy:
You'll need to access the values:
Sth like: tbl['nadcs'].nda[ii] = fcio.nadcs

jasondet:
close. they are scalars not arrays, so you need `value` instead of `nda`:
```
>>> from pygama import lh5
>>> store = lh5.Store()
>>> fc, n = store.read_object('fcio_config', 'file.h5')
>>> fc['nadcs'].value
6                                 # may vary 
```

Alternatively, you can access them directly with h5py:
```
>>> import h5py
>>> f = h5py.File('file.h5', 'r')
>>> f['fcio_config/nadcs'][()]
6                                 # may vary 
```

Kermaidy:
Right, they are scalars but I meant, you need to get the values from FCIO (see e.g.: `numtraces = fcio.numtraces`) which I think you do not at the moment. Or I'm missing something? I just expected smth like `nadcs = fcio.nadcs`

jasondet:
Oh -- the config fields were already being stored to memory here:

https://github.com/legend-exp/pygama/blob/dev/pygama/io/fcdaq.py#L386

which calls

https://github.com/legend-exp/pygama/blob/dev/pygama/io/fcdaq.py#L141

PR #136, opened 2021-02-11
need to check type as string

jasondet:


wisecg:
Looks ok.  On to v0.5!

PR #139, opened 2021-02-12
fix bug found by mariia

wisecg:


jasondet:
@wisecg -- for the "version" setting -- slightly above that in setup.py we are using git.Repo to populate the git.py file in pygama/pygama:
https://github.com/legend-exp/pygama/blob/master/setup.py#L87
It looks like we should be able to use a function like repo.tag to automatically populate that field. Can you try that instead?

The fix in store.py looks good, thx.

iguinn:
I'm actually not so sure using git.Repo is a good idea for populating that field. If you are using master, it will not use the tag, or if you are on a floating head. In addition, if someone is installing without git for some odd reason (e.g. if someone were given a zip of the repo or something), they will not have access to the version number.

jasondet:
ok I'm going to merge anyway, and let's continue this discussion in an issue

PR #147, opened 2021-02-22
[WIP] Pulse Injector Processor

SeanSullivan1:
Added pulse injector consisting of two functions to add exponential or sigmoid based pulses to waveforms using start time, rise time, amplitude and decay constant of pulse.

### DIFF
--- a/pygama/dsp/_processors/pulse_injector.py
+++ b/pygama/dsp/_processors/pulse_injector.py
@@ -1,0 +1,28 @@
+import numpy as np
+from numba import guvectorize
+from math import log,exp
+
+@guvectorize(["void(float32[:], float32, float32, float32, float32, float32[:])",
+              "void(float64[:], float64, float64, float64, float64, float64[:])"],
+              "(n),(), (), (), ()->(n)", nopython=True, cache=True)
+def inject_sig_pulse(wf_in,t0,rise,amp,tau,wf_out):
+    """
+    Inject sigmoid pulse into existing waveform to simulate pileup
+    """    
+    wf_out[:] = wf_in[:]
+    for T in range(len(wf_out)):
+        wf_out[T] = wf_out[T]+ (amp/(1+exp((-1/rise)*(T-(t0+rise/2))))*exp(-(1/tau)*(T-t0)))
+
+@guvectorize(["void(float32[:], float32, float32, float32, float32, float32[:])",
+              "void(float64[:], float64, float64, float64, float64, float64[:])"],
+              "(n),(), (), () , ()->(n)", nopython=True, cache=True)     
+def inject_exp_pulse(wf_in,t0,rise,amp,tau,wf_out):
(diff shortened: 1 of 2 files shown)


jasondet:
@mmatteo is this still WIP? if so I would prefer to close the PR and reopen it when its ready

mmatteo:
@jasondet, the code is working but not reviewed yet. I used WIP to make clear that is not ready to be merged

pygama/dsp/_processors/pulse_injector.py
```diff
@@ -0,0 +1,28 @@
...
+@guvectorize(["void(float32[:], float32, float32, float32, float32, float32[:])",
+              "void(float64[:], float64, float64, float64, float64, float64[:])"],
+              "(n),(), (), (), ()->(n)", nopython=True, cache=True)
+def inject_sig_pulse(wf_in,t0,rise,amp,tau,wf_out):
+    """
+    Inject sigmoid pulse into existing waveform to simulate pileup
+    """    
+    wf_out[:] = wf_in[:]
+    for T in range(len(wf_out)):
+        wf_out[T] = wf_out[T]+ (amp/(1+exp((-1/rise)*(T-(t0+rise/2))))*exp(-(1/tau)*(T-t0)))
```
mmatteo, line 14: @SeanSullivan1, I think line 12 is not needed. You can use directly wf_in at line 14. Same in the other function

pygama/dsp/_processors/pulse_injector.py
```diff
@@ -0,0 +1,28 @@
...
+
+@guvectorize(["void(float32[:], float32, float32, float32, float32, float32[:])",
+              "void(float64[:], float64, float64, float64, float64, float64[:])"],
+              "(n),(), (), (), ()->(n)", nopython=True, cache=True)
+def inject_sig_pulse(wf_in,t0,rise,amp,tau,wf_out):
+    """
+    Inject sigmoid pulse into existing waveform to simulate pileup
+    """    
+    wf_out[:] = wf_in[:]
+    for T in range(len(wf_out)):
```
mmatteo, line 13: why don't you start the loop from T=t0?

mmatteo (changes requested):

pygama/dsp/_processors/pulse_injector.py
SeanSullivan1: For injecting pre trigger pileups I use a negative time, so starting from t0 would start to use negative indices for adding the pulse to the existing waveform and would produce incorrect results. I Could use an if statement to check the current index is positive before adding the pulses at that point, or maybe there's another work around. Not sure what the preferred solution would be here.

wisecg:
@SeanSullivan1 Is this still under development?  If so, I suggest you close this PR, pull the recent changes we've made from the `dev` branch, and re-submit.

PR #148, opened 2021-02-24
Bug fixes to lpgta/processing.py and lpgta/setup.py.

sweigart:
I fixed a number of bugs in processing.py and setup.py, which had prevented them from being run.  I also added back in the --bl and --bw options in processing.py, which had been previously removed by accident.


wisecg:
Thanks for getting this back in shape, David!

PR #151, opened 2021-02-25
1D Gaussian Filter

SamuelBorden:
Added one-dimensional Gaussian filter for dsp based on scipy's gaussian_filter1d function. The function takes in a waveform, and both a sigma and a truncation float (to compute the Gaussian kernel), and then returns waveform out.

### DIFF
--- a/pygama/dsp/_processors/gaussian_filter1d.py
+++ b/pygama/dsp/_processors/gaussian_filter1d.py
@@ -1,0 +1,104 @@
+# Copyright (C) 2003-2005 Peter J. Verveer
+#
+# Redistribution and use in source and binary forms, with or without
+# modification, are permitted provided that the following conditions
+# are met:
+#
+# 1. Redistributions of source code must retain the above copyright
+#    notice, this list of conditions and the following disclaimer.
+#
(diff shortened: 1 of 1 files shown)


iguinn (changes requested):
Hey Samuel,
As currently written, this will recalculate the Gaussian kernel every time the processor is run, which we would like to avoid since this will slow things down a lot. The trick is to define a factory function which generates the kernel, and then generates the @guvectorize processor from inside the function, and returns that.

In pseudo code, that would look like:
`def gaussian_filter1d(sigma, truncate):
    ...do stuff to calculate kernel
    #guvectorize(...)
    def gaussian_filter1d_out(wf_in, wf_out)
        convolve wf_in with kernel
    return gaussian_filter1d_out
`
The kernel will then be hard-coded into the numba-compiled function gaussian_filter1d_out so that it doesn't have to be recalculated.

For an example of how to use this, the cusp_filter uses this technique. It should be noted that right now Sean and I are trying to debug a problem with actually including it in an analysis.
Ian

iguinn (approved):

PR #153, opened 2021-03-01
Improvements related to lpgta/processing.py.

sweigart:
I made some general improvements related to lpgta/processing.py:

- The `overwrite` argument in raw_to_dsp.py now controls whether or not to delete the dsp file before processing the raw file.  Previously, this argument was unused.
- The --date option in lpgta/processing.py is now AND-ed with the --q option in order to come up with the query that is passed to the `load_datagroup` function.  Previously, the --date option would override the --q option.
- The `load_datagroup` function is now run whenever the --d2r and/or --r2d option is used even when the --dg option is not used.  Previously, this was an implicit requirement.  The --dg option is retained because it is useful for testing out different queries and seeing what files would be processed without actually processing them.
- The arguments used in processing.py is now printed out when `verbose` is True.  This will allow us to record the exact configuration used to a log file when processing the data.

### DIFF
--- a/experiments/lpgta/processing.py
+++ b/experiments/lpgta/processing.py
@@ -31,8 +31,8 @@ def main():
 
     # declare datagroup
     arg('--dg', action=st, help='load datagroup')
-    arg('--q', type=str, help='datagroup query')
-    arg('--date', type=str, help='date query')
+    arg('--q', type=str, help='datagroup query (AND-ed with the date query)')
+    arg('--date', type=str, help='date query (AND-ed with the datagroup query)')
 
     # routines
     arg('--d2r', action=st, help='run daq_to_raw')
@@ -47,6 +47,10 @@ def main():
     arg('-c', '--cwd', action=st, help="save output to current directory")
     args = par.parse_args()
 
+    # print out the arguments for log files
+    if args.verbose:
+        print('Arguments:', args)
+
     # -- set options --
     nwfs = args.nwfs if args.nwfs is not None else np.inf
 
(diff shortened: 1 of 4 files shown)


sweigart:
The second commit to this pull request was mistakenly added and is unrelated to the first.  I've now learned how to prevent this mistake from happening again.  If it's acceptable, it'd be great to have both commits be pulled in.  Otherwise, I can try to undo it and make a separate pull request.

This new commit's changes to fcdaq.py allow for the output file names that are not actually saved to be printed out.  This is a helpful message for the user who might have been expecting an output file.  More specifically, it helps explain whether no file was saved because there were no data to save or because there was an error in the saving process.

mmatteo:
Hi @iguinn @sweigart, as part of the currently ongoing pygama refactory, the part of the code handling the IO will be decoupled from the part of the code processing in-memory data. @jasondet can comment better here, but I expect that we will get rid of the overwrite option in raw_to_dsp and handle it as part of the data flow manager

jasondet:
proper overwrite (at the file as well as dataset level) is now [implemented but untested in LH5Store in the refactor branch on my fork](https://github.com/jasondet/pygama/commit/13cbc89)

raw_to_dsp will renamed something like build_dsp(); the guts in pygama will be a table_in - to - table_out function, and we will at least have a default function that can run from file-to-file, although some experiments (perhaps LEGEND) will need their own implementation for more experiment-specific data handling and might be done within data flow management directly.

jasondet:
@sweigart these changes look good to me. I'm going to merge them.

iguinn:
I'm going to move some of this conversation over to an issue

Issue #130, opened 2021-02-08
Exception Handling in ProcessingChain


mmatteo:
Can the ProcessingChain handle exceptions raised by the processors? 

If a processor fails for a waveform, all processors depending on it should be skipped. I would also store a failedFlag in the dsp tier to summarize the exit status of all processors.

What do you think?

iguinn:
Hey Matteo,
Thanks for raising this issue! This is still something that needs work. Right now, an exception should stop the full run, which will almost never be the desired behavior (perhaps if there is a case where a processor would just always fail this would make sense?). I believe that numba will properly throw the python exceptions if they occur in processors, so the solution is just going to be to add a try-catch block in the right place in the code. One important thing to keep in mind is that since waveforms are processed in blocks, it is not possible to skip just a single waveform; it would instead skip all waveforms in a block!

For the failedFlag, would this just be one bool for each processor so that at the end we can print out a list of failures?

mmatteo:
@iguinn, thanks for the explanation. PR #129 contains an example of a processor that will fail for low energy events. The processor is supposed to compute the log of the samples and, before running, it checks that the value of all samples is positive. @ggmarshall's implementation returns a np.nan waveform if there are negative sample. Is this what we want?

jasondet:
I would prefer a different approach, which is to require that no processor should raise an exception for valid input data. It's better to keep problems from happening in the first place than to take a more lax approach and just try to catch them. Is the functionality of PR #129 worth jettisoning guaranteed validity?

mmatteo:
@jasondet, I think it is normal that a processor might not be able to return a meaningful result for some classes of events. For instance, most of the current processor will have trouble with low energy events. The leading-edge trigger cannot find a trigger if there is no pulse and, in this case, any module depending on it -- e.g. the fixed time pickoff -- should not run.

jasondet:
@mmatteo again, I would rather have fully populated output tables and check for strange values afterwards rather than have tables with values missing or rows skipped for some of the parameters. 

Or maybe I don't understand your recommendation. Considering your example -- if we are running a fixed time pickoff and expecting its output to be in the output table, what goes in the output table to hold the place for that row / event if we have decided to "turn off" FTP because of an error in one of its dependencies for that event's waveform?

mmatteo:
[...]

PR #158, opened 2021-03-24
Added radford_fwhm to peak_fitting.py and ProcessingChain improvements

iguinn:
Numerically solve for FWHM


iguinn:
This pull request is not much more than a numerical FWHM solution. Now, this also includes some improvements to ProcessingChain, including:
- error handling: Added pygama.dsp.errors, which defines DSPFatal and ProcessingChainError. DSPFatal is to be thrown by processors when a problem should abort processing. ProcessingChainError is there to clarify when a problem is raised during the setup of a processing chain. Added some error handling for processors, which will add some clarity about what event caused the problem
- Make type signatures less strict. Before, we required an exact match between the type signatures of a processor and those of the inputs. However, numpy ufuncs are fully capable of automatically casting arrays, so this was unnecessarily restrictive. Now we will only require that a variable can be safely converted into a type usable by a processor
- Got rid of the need for the pesky prereqs lines in the dsp json files. Now, build_processing_chain is capable of automatically figuring these out. If it messes up, it is still an option to provide prereqs in the json file.
- Clarified/improved error handling in both ProcessingChain and build_processing_chain
- Changes to processors:
   - fixed_time_pickoff now accepts a float for the time and interpolates between samples
   - trap_filter and trap_norm now raise exceptions when the trap is too long

wisecg:
We discussed this quite a bit in this issue: https://github.com/legend-exp/pygama/issues/130

PR #163, opened 2021-04-19
Processor skeleton and coding style

mmatteo:
This is a proposal for a common style/structure of all processors. Once we agree on it, I will go ahead and refactor the existing processors. A few topics for discussion:

1. @iguinn, what about using a single set of 64bit variables in the guvectorize decorator?  This should improve the readability of the code and newcomers. To my understanding, using 64 bits will only impact memory usage, which is not an issue in our case. 
2. the initialization of the ouput parameters to NaN could in principle be moved to the processing chain. I don't mink keeping it here, but it will lead to code duplication
3. pygama_abort is a place holder until @iguinn provides the new function
4. do you agree on using a range-based loop as the default method for looping over the waveform?


pygama/dsp/_processors/template.py
```diff
@@ -0,0 +1,77 @@
...
+import numpy as np
+from numba import guvectorize
+
+# 2) Provide instructions to numba
+# 
+# Documentation about numba guvectorize decorator:
+# https://numba.pydata.org/numba-doc/latest/user/vectorize.html#the-guvectorize-decorator
+#
+# Notes:
+# - use `nopython=True` and `cache=True`         (@ian, is this correct?)
```
iguinn, line 12: In most cases this is correct. The exceptions are:
- if the processor relies on a command that numba does not recognize, then you want to use `objectmode=True`. For example, if you are using something not on this list of accepted numpy operations: https://numba.pydata.org/numba-doc/dev/reference/numpysupported.html, or if you are wrapping a function from another module
- `cache=True` does not apply if you are using a factory function to apply some sort of initialization to the processor

pygama/dsp/_processors/template.py
```diff
@@ -0,0 +1,77 @@
...
+# - use 64-bit variables, e.g. int64 and float64 (@all, this is up for discussion)
```
iguinn, line 13: There are performance benefits to using 32-bit variables for processors that take advantage of SIMD vectorized operations, since more 32-bit values than 64-bit values can be processed at once

pygama/dsp/_processors/template.py
```diff
@@ -0,0 +1,77 @@
...
+# - use [:] for al output parameters             (@ian, is this correct?)
```
iguinn, line 14: Yes this is correct. @guvectorize should fail if you do not do this

pygama/dsp/_processors/template.py
```diff
@@ -0,0 +1,77 @@
...
+             "(n),(),()->()", nopython=True, cache=True)
+
+# 3) Define the processor interface
+#
+# Notes: (@all, this is up for discussion)
+# - add the `_in`/`_out` suffix to the name of the input/output variables
+# - use w_ for waveforms, t_ for indexes, a_ for amplitudes 
+# - use underscore casing for the name of the processor and variables, e.g.
+#   a_trap_energy_in or t_trigger_in
+#
```
iguinn, line 26: I agree with all of these

pygama/dsp/_processors/template.py
```diff
@@ -0,0 +1,77 @@
...
+def the_processor_template(w_in, t_in, a_in, w_out, t_out):
+
+    # 4) Document the algorithm
+    #
+    """
+    Add here a complete description of what the processor does, including the
+    meaning of input and output variables
+    """
+
+    # 5) Initialize output parameters (@iann, will this be done automatically by the processing chain?)
```
iguinn, line 36: The problem with trying to initialize the output parameters is that a lot of unexpected behavior will still overwrite these values. For example, if you feed fixed_time_pickoff an index that is out of range, but not by enough to cause a segfault, it will overwrite it with a nonsense value. I think it is a better idea to explicitly handle the cases where we want to use a NaN value within the processor

pygama/dsp/_processors/template.py
```diff
@@ -0,0 +1,77 @@
...
+    #   output parameters to propagate the failure throughout the processing chain.
+    # - in range checks. Check if indexes are within 0 and len(waveform),
+    #   amplitudes are positive, etc. A failure of this check implies errors in
+    #   the dsp json config file. Abort the analysis immediately.
+    #
+    if (np.isnan(w_in).any() or t_in == np.nan or a_in == np.nan):
+        return
+
+    if (not  0 <= t_in < len(w_in) or not a_in >= 0):
+        pygama_abort();
```
iguinn, line 62: This will be done by raising an exception. I will add a try-catch block around the processors to handle these by printing out useful information that can help with debugging, and is not always available to numba functions

iguinn:
> @iguinn, what about using a single set of 64bit variables in the guvectorize decorator? This should improve the readability of the code and newcomers. To my understanding, using 64 bits will only impact memory usage, which is not an issue in our case.
For processors that use SIMD vectorized processor instructions, there is an advantage to using 32 bits since more 32-bit values can fit in the processor register than 64-bit values. We also will very rarely need 64-bits for these. I agree with the goal of trying to simplify these, but I think it might be worth it to keep 32- and 64-bit versions (on the other hand, I included some int waveforms that are totally unnecessary and can be removed).

Another time we might want to include multiple signatures is for time values. Right now I have defined them all as ints, but int values do not allow NaN's, so I think we want to use floats for a lot of the timepoints. On the other hand, we use `argmax` which returns ints, so we may want to include support for both (there is no strong reason not to use 64 bits here, though 32 would probably be sufficient). For processors that take times in, this would require us to do some interpolation.

> the initialization of the ouput parameters to NaN could in principle be moved to the processing chain. I don't mink keeping it here, but it will lead to code duplication
See my notes in the change log.

> pygama_abort is a place holder until @iguinn provides the new function
This will end up being `raise DSPError('description')`. By using a try-catch block, I can add additional useful information (e.g. where in the file was it thrown, and what was the actual function call).

I am struggling a bit to find a suitable solution for warnings, where we want to continue the processing but record some sort of message. Since these are vectorized, if we raise an exception and handle it specially, this will also abort any waveforms following the one that caused the problem. Another option is to just look for NaN's, but it is harder to provide useful information like a description that way. Unfortunately, numba makes it difficult to communicate any information outside of the function.

> do you agree on using a range-based loop as the default method for looping over the waveform?
Yes, although in certain situations using numpy-style vector operations also makes sense (especially if you are not in nopython mode, which makes loops extremely slow!)

pygama/dsp/_processors/template.py
mmatteo: I see. Do you think it is worth adding these comments to the template?

pygama/dsp/_processors/template.py
mmatteo: if there is a clear advantage, I am fine with declaring multiple set of input variables. However, I would try to keep the number of sets limited, and define a clear standard on which sets should be declared. It is easy to have one set of 32-bit variables and another with 64-bit variables. I would however avoid sets with a mixture of 32- and 64-bit variables as it can be confusing to decide which is what.

Btw, which declaration is used when multiple are defined? I don't like that the user does not have control over this.

pygama/dsp/_processors/template.py
mmatteo: Alright, if everybody is fine with that I will implement initialize the outputs at the beginning of each processor

pygama/dsp/_processors/template.py
mmatteo: Is this already available?

pygama/dsp/_processors/template.py
iguinn: Yeah, I agree that we should have some specification that prevents the number of signatures from blowing up. For waveforms and amplitude values, I think dividing between 32 and 64 bit makes sense. For timing values, I agree that floats are preferable, but there is (at least) one case where we will be getting int32s, which is if we use the argmax function from numpy.

When multiple declarations are available, it will use the types that match what is input. When calling it directly, it will cast values as needed, if the direct types aren't available; in the ProcessingChain class, I included code to find the right type signature. This was necessary so that the code can deduce how to create new internal memory buffers of the right type and size. If it encounters data types that are not compatible with any of the signatures, it will raise an exception. Unfortunately, the way I wrote it does not allow for casting to other types, so if it sees an int32, it won't cast up to float64 automatically.

mmatteo: I added two declarations, one for 32-bit variables and the other one for 64-bit variables. I propose not to support the usage of integers as we need the NaN value to handle the expectations. Would it be possible to implement a check in the processing chain and cast any integer into a floa?

PR #164, opened 2021-05-21
Processor Refactoring

ggmarshall:


mmatteo:
@jasondet @wisecg @iguinn, can you please review this PR?

jasondet:
@iguinn please review when you are back from vacation.
@sweigart if you have a chance please take a look as well.

iguinn:
These changes all look good to me. How much have you tested these on real data?

ggmarshall:
I've checked they give the same results as the old processors on a set of 1000 waveforms, since then I have processed all the HADES data (around 25 detectors) using these processors.

iguinn:
Ok, sounds good. I'll give it a quick try on PGT data, just to check for exceptions and if any NaNs show up unexpectedly.

sweigart:
This update looks nice.  I just have a couple of recommendations and a question:

- fixed_time_pickoff.py:  There’s a "(c) 2021 GitHub, Inc." in the last line, which I think should throw a syntax error if run.  I recommend that this line is deleted.
- mean_stdev.py:  This processor was deleted, but I currently use it in my analysis.  I recommend keeping it if possible.
- time_point_frac.py:  This processor was deleted, which has been typically used in the "tp_90" DSP output variable and the like.  Is the idea to switch such variables to use the updated time_point_thres.py instead?

ggmarshall:
Hi David, good catch on the ftp, I'll fix that now.
Regarding mean_stdev and time point frac both of these processors functionality are covered by other processors. For time point frac the threshold of the time point thresh can simply be defined in the processing chain as some fraction of a variable e.g. 0.9*energy. Mean_stdev is covered by the linear_slope_fit processor.

iguinn:
I can confirm that the copyright David caught causes a Syntax Error when processing. I am just using raw2dsp.py with a default config file, but it also includes the mean_stdev and timepoint_frac processors, so we will want to reinclude these as well.

sweigart:
Thanks, George, for explaining the processor overlap.  I like having time_point_thres taking the place of time_point_frac since it's conceptually doing the same thing.  If desired, I'm also fine with switching my analysis code to use linear_slope_fit in place of mean_stdev (and just ignore the slope and intercept output variables).

pygama/dsp/_processors/fixed_time_pickoff.py
```diff
@@ -1,20 +1,44 @@
...
-        value_out[0] = wf_in[t]*(1-frac) + wf_in[t+1]*frac
+    a_out[0] = np.nan
+
+    if (np.isnan(w_in).any() or np.isnan(t_in)):
+        return
+    
+    if (not int(t_in) in range(len(w_in))):
+        return
+  
+    if (not np.floor(t_in)==t_in):
```
iguinn, line 41: This line is throwing errors for me, since the time_point_finder returns a float

pygama/dsp/_processors/time_point_thresh.py
```diff
@@ -1,21 +1,59 @@
...
+import math
+from pygama.dsp.errors import DSPFatal
 
-@guvectorize(["void(float32[:], float32, int32, int32[:])",
-              "void(float64[:], float64, int32, int32[:])",
-              "void(int32[:], int32, int32, int32[:])",
-              "void(int64[:], int64, int32, int32[:])"],
-             "(n),(),()->()", nopython=True, cache=True)
-def time_point_thresh(wf_in, threshold, tp_max, tp_out):
+@guvectorize(["void(float32[:], float32, float32, int32 ,float32[:])",
```
iguinn, line 6: It turns out that using float32s for time values will cause conflicts with time values passed as int32s. This happened to me using the output of argmax as t_start here. A solution is to use float64s for time values, since int32 can safely convert to float64

iguinn:
In order to maintain backwards compatibility with the processors, I think we should keep the old ones, but add a deprecation warning. The reason is that multiple DSP config files are floating around that will break after we pull this in, and this will, in principle, give people a chance to fix them before we remove these altogether at some later date. I have submitted a pull request to @ggmarshall's fork with code added to do this here: https://github.com/ggmarshall/pygama/pull/4

mmatteo:
@iguinn, I would prefer to fully break backward compatibility and force the users to update their DSP config file. The reason is that the behaviour of the new processors -- e.g. NaN return values and exception handling -- can already affect the results of the analysis, and users might not realize it immediately.  Let's push them to work with the stable pygama tagged version as long as the new release is ready. The new release should come with a detailed changelog of the user interfaces.  I would even consider adding a  pygama-version filed in the JSON files. This could help to handle future developments

pygama/dsp/_processors/fixed_time_pickoff.py
ggmarshall: I've removed this condition as it also causes problems when using the values obtained by the optimizer

ggmarshall:
Hi @iguinn, I've modified the default config to now use the new processors instead.

iguinn:
@ggmarshall Thanks for the changes. I've successfully been able to process a PGT file, so I'm happy with the changes
@mmatteo I agree with that reasoning

jasondet:
It looks to me like all issues have been addressed here. I'll wait another day or so for comments then merge if there is no reason to wait.

mmatteo:
Thanks @jasondet, I would merge this and continue the revision of the processors (if needed) in a separate PR

PR #165, opened 2021-05-26
Store.read_object will now read a list of file names from any iterable object

iguinn:
Most importantly, this will work for pandas objects, which is nice for use with datagroup


iguinn:
Added a new commit. Now setup requires h5py version >= 3.2.0 to include this bug fix from our own @jasondet https://github.com/h5py/h5py/issues/1792

iguinn:
Added a bug fix to the Gretina4M digitizer, caused by a change in python where at some point dict.items went from returning a list of items to an iterator object

jasondet (approved):
Looks good!

PR #169, opened 2021-07-12
Added multi-t-filter

SamuelBorden:
First draft of multi-triggering-filter that takes in a waveform and returns a list of tp0s of multiple peaks within a waveform. It is a vectorized function that has to call special njit'ed versions of time_point_thresh and get_i_local_extrema. I couldn't get it to operate in place on the tp0_list array, so it has to be fed an array of the same length (called "arbitrary_list") to tell Numba what length array to return.


iguinn:
Hey Sam, thanks for writing these, they will be extremely useful! In addition to its utility for LAr data as you showed, we can also use them for pileup in Ge events in the future. Here are a few general comments:

1. We want to make sure to follow the template outlined by Matteo and George here: https://github.com/legend-exp/pygama/blob/dev/pygama/dsp/_processors/template.py. A couple places that you'll want to change include: 
     - The variable naming scheme; eg. `change` data to `w_in`. You have a couple of new cases here that are worth discussing. First, for the length of variable length vectors, I like `n` as a prefix, so `n_max` and `n_min` should be good. Second, for the vectors themselves, we may want to consider something like `v` as a prefix, so for example `max_list` might become `v_max`. Alternatively, since this represents an amplitude we could stick to the `a` convention, or maybe do something like `va` for vector of amplitudes, `vt` for vector of timepoints, etc. I'm interested in @mmatteo and @jasondet's thoughts on this.
     - Use `_in` and `_out` suffixes
     - When throwing exceptions, use the DSPFatal exception

2. We want to try to use guvectorize for everything, rather than njit. The reason for this is that guvectorize produces a ufunc object that has information used by ProcessingChain to pre-allocate all of the memory that gets used. My understanding is that you used njit because numba won't allow you to use guvectorized functions within guvectorized functions in nopython mode (which is really frustrating).

Even so, it would be nice to have the multi min/max function, for example, as a guvectorize in case we want to use it for other purposes. My suggestion here would be to write multi_triggering_filter using object mode, and then just call a sequence of guvectorized functions. We can't quite convert this directly, because it contains some for loops, and for those we prefer nopython mode, so I'll put some inline comments by the function with some suggestions.

Ian

pygama/dsp/_processors/multi_t_filter.py
```diff
@@ -0,0 +1,169 @@
...
+    tp_out = 0
+    return tp_out
+    
+
+
+@guvectorize(["void(float32[:], float32[:],float32[:], float32, float32[:])",
+              "void(float64[:], float64[:], float64[:], float64, float64[:])",
+              "void(int32[:], int32[:], int32[:], int32, int32[:])",
+              "void(int64[:], int64[:], int64[:], int64, int64[:])"],
+             "(n),(m),(),()->(m)", nopython=True, cache=True)
```
iguinn, line 110: As I said in my longer comment, we may want to write this using `forceobj=True` instead of nopython mode so that we can call a series of guvectorized functions

pygama/dsp/_processors/multi_t_filter.py
```diff
@@ -0,0 +1,169 @@
...
+    n_max = np.array([0],dtype=np.int32)
+    n_min = np.array([0],dtype=np.int32)
+    flag = np.array([0],dtype=np.int32)
+
+    # call njit_get_i_local_extrema to get locations of the maxima and minima within a waveform
+    
+    njit_get_i_local_extrema(wf_in, delta,max_list,min_list,n_max,n_min,flag)
+    
+    # Go through the list of maxima, calling njit_time_point_thresh on each maximum (ignoring the nan padding)
+    for i in range(int(n_max[0])):
```
iguinn, line 149: To avoid a loop here, you should be able to use the refactored time_point_thresh that George wrote. Ufuncs do broadcasting of dimensions, meaning it will figure out that there is an extra dimension to loop over and automatically execute that loop for you. In addition, George added NaN checks to that function, so it will work the same way as written here.

pygama/dsp/_processors/multi_t_filter.py
```diff
@@ -0,0 +1,169 @@
...
+    """ 
+    time_point_thresh has issues with afterpulsing in waveforms that causes  
+    an aferpulse peak's tp0 to be sent to 0 or the same index as the tp0 for the first pulse.
+    This only happens when the relative minimum between the first pulse and 
+    the afterpulse is greater than the threshold. So, we sweep through the array again 
+    to ensure there are no duplicate indices. If there are duplicate indicies caused by a
+    misidentified tp0 of an afterpulse, we replace its index by that of the corresponding minimum
+    found using the njit_get_i_local_extrema function.
+    """
+
```
iguinn, line 162: I would suggest turning this next block into a new guvectorized processor called something like `remove_duplicates` or something like that

pygama/dsp/_processors/multi_t_filter.py
```diff
@@ -0,0 +1,169 @@
...
+    k=0
+    for index, i in enumerate(tp0_list):
+        for index2, j in enumerate(tp0_list[index+1:]):
+            if i == j:
+                tp0_list[index2+index+1] = min_list[k]
+        k+=1 # this makes sure that the index of the misidentified afterpulse tp0 is replaced with the correct corresponding minimum
+
```
iguinn, line 169: In the end, what I think this function should end up looking like as an objectmode function is:
def multi_triggering_thresh(...):
    get_i_local_extrema(...)
    timepoint_thresh(...)
    remove_duplicates(...)

This way, we avoid any python loops that are slow in objectmode

SamuelBorden:
Thanks for the feedback Ian! I changed the multi_t_filter to use object mode. I also renamed and split the get_multi_local_extrema function to its own file, and then imported both that and the refactored time_point_thresh into multi_t_filter.  As you recommended, I also created a guvectorized function for removing duplicates (and also changed the for loop in it to avoid using enumerate, per the template suggestions).

pygama/dsp/_processors/get_multi_local_extrema.py
```diff
@@ -0,0 +1,93 @@
...
+    
+    n_max_out[0] = 0
+    n_min_out[0] = 0
+    flag_out[0] = 0
+    
+
+    # sanity checks
+    w_in, delta2 = np.asarray(w_in), np.asarray(delta_in)
+    if delta2.ndim != 0:
+        raise DSPFatal("Input argument delta_in must be a scalar")
```
iguinn, line 54: I think that the logic that handles numpy/numba broadcasting should make this check unnecessary, since the value has dimension () in the signature. On the other hand, if you've run into a situation where this has happened, please say so, and in that case we will want to add this check other places!

pygama/dsp/_processors/multi_t_filter.py
```diff
@@ -0,0 +1,93 @@
...
+             "(m),(m)", nopython=True, cache=True)
+def remove_duplicates(t_out,vt_min_out):
+    """ 
+    time_point_thresh has issues with afterpulsing in waveforms that causes  
+    an aferpulse peak's tp0 to be sent to 0 or the same index as the tp0 for the first pulse.
+    This only happens when the relative minimum between the first pulse and 
+    the afterpulse is greater than the threshold. So, we sweep through the array again 
+    to ensure there are no duplicate indices. If there are duplicate indicies caused by a
+    misidentified tp0 of an afterpulse, we replace its index by that of the corresponding minimum
+    found using the njit_get_i_local_extrema function.
```
iguinn, line 21: njit_get_i_local_extrema has been renamed, so this needs to be changed too. I'd also double check the doc-strings in case there's anything else that needs to be changed

iguinn:
Hey Sam,
Looks good, aside from the couple of minor comments.
Since this is our first processor that is explicitly using a variable-width vector output, I also want to double check with @mmatteo that we're all happy with the convention adopted here of preceding the vector variables with v, so a vector of time variables becomes `vt_...`, and preceding the variables that represent the size of the vectors with `n_...`.
Ian

pygama/dsp/_processors/get_multi_local_extrema.py
```diff
@@ -0,0 +1,93 @@
+import numpy as np
+from numba import guvectorize, njit
+from pygama.dsp.errors import DSPFatal
+
+
+
+@guvectorize(["void(float32[:], float32, float32[:], float32[:], float32[:], float32[:], float32[:], int32[:])",
+              "void(float64[:], float64, float64[:], float64[:], float64[:],float64[:], float64[:], int32[:])",
```
mmatteo, line 8: Our convention is to use only floats, even if the parameters are integers. The reason is that we use NaN return values to track failures and integers cannot be NaN. I suggest changing the int32 into a float32 in the first line and into a float64 in the second line

pygama/dsp/_processors/get_multi_local_extrema.py
```diff
@@ -0,0 +1,93 @@
...
+              "void(int32[:], int32, int32[:], int32[:], int32[:], int32[:], int32[:], int32[:])",
+              "void(int64[:], int64, int64[:], int64[:], int64[:], int64[:], int64[:], int32[:])"],
```
mmatteo, line 10: can we remove these two declarations?

pygama/dsp/_processors/get_multi_local_extrema.py
```diff
@@ -0,0 +1,93 @@
...
+
+    # prepare output 
+    #padded with nans or zeros depending on what we want
+    vt_max_out[:]= np.nan
+    vt_min_out[:] = np.nan
+
+    
+    n_max_out[0] = 0
+    n_min_out[0] = 0
+    flag_out[0] = 0
```
mmatteo, line 48: I would set all these output parameters to nan and return 0 only if the processor succeeds

pygama/dsp/_processors/get_multi_local_extrema.py
mmatteo: I would not use functions that modify the input waveform. We want all inputs to be treated as const objects. All operations should be done on tmp or output objects

mmatteo: I would add the nan checks here: see https://github.com/legend-exp/pygama/blob/5e261e69eec1610431fe2eda42d994ba58d299ea/pygama/dsp/_processors/template.py#L49

pygama/dsp/_processors/multi_t_filter.py
```diff
@@ -0,0 +1,93 @@
...
+from numba import guvectorize, njit
+from time_point_thresh import time_point_thresh
+from get_multi_local_extrema import get_multi_local_extrema
+
+
+    
+@guvectorize(["void(float32[:],float32[:])",
+              "void(float64[:],float64[:])",
+              "void(int32[:],int32[:])",
+              "void(int64[:],int64[:])"],
```
mmatteo, line 11: let's remove the last two declarations

pygama/dsp/_processors/multi_t_filter.py
```diff
@@ -0,0 +1,93 @@
...
+        k+=1 # this makes sure that the index of the misidentified afterpulse tp0 is replaced with the correct corresponding minimum
+    if not np.isnan(t_out[0]): # makes sure that the first maximum found isn't the start of the waveform
+        if int(t_out[0]) == 0: 
+            t_out[:] = np.append(t_out[1:],np.nan)
+
+
+@guvectorize(["void(float32[:], float32[:],float32[:], float32, float32[:])",
+              "void(float64[:], float64[:], float64[:], float64, float64[:])",
+              "void(int32[:], int32[:], int32[:], int32, int32[:])",
+              "void(int64[:], int64[:], int64[:], int64, int64[:])"],
```
mmatteo, line 43: remove last two declarations

pygama/dsp/_processors/get_multi_local_extrema.py
```diff
@@ -0,0 +1,93 @@
...
+            # if the sample is more than the current min by more than delta,
+            # declare the previous one a minimum, then set this as the new "max"
+            if w_in[i] > w_in[imin] + delta_in and int(n_min_out[0]) < int(len(vt_min_out)):
+                vt_min_out[int(n_min_out[0])] = imin
+                n_min_out[0] += 1
+                imax = i
+                find_max = True
+    if n_max_out[0] != 0:
+        flag_out[0] = 1
+
```
mmatteo, line 86: else:
   flag_out[0] = 0

pygama/dsp/_processors/multi_t_filter.py
```diff
@@ -0,0 +1,93 @@
...
+    to ensure there are no duplicate indices. If there are duplicate indicies caused by a
+    misidentified tp0 of an afterpulse, we replace its index by that of the corresponding minimum
+    found using the njit_get_i_local_extrema function.
+    ----------
+    t_out : array-like
+        the array of indices that we want to remove duplicates from 
+    vt_min_out : array-like
+        list of indicies of minima that we want to replace duplicates in t_out with
+    """
+
```
mmatteo, line 28: add initializations and checks, see points 5 and 6: https://github.com/legend-exp/pygama/blob/5e261e69eec1610431fe2eda42d994ba58d299ea/pygama/dsp/_processors/template.py#L39

pygama/dsp/_processors/multi_t_filter.py
```diff
@@ -0,0 +1,93 @@
...
+    """
+     
+    # initialize arrays, padded with the elements we want
+    
+    t_out[:] = np.nan 
+    vt_max_out= np.zeros_like(t_arbitrary_in, dtype=np.float32)
+    vt_min_out = np.zeros_like(t_arbitrary_in,dtype=np.float32)
+    n_max_out = np.array([0],dtype=np.float32)
+    n_min_out = np.array([0],dtype=np.float32)
+    flag_out = np.array([0],dtype=np.float32)
```
mmatteo, line 81: can we set all of these to Nan?

pygama/dsp/_processors/multi_t_filter.py
```diff
@@ -0,0 +1,93 @@
...
+
+    # call get_i_local_extrema to get locations of the maxima and minima within a waveform
+    
+    get_multi_local_extrema(w_in, delta_in,t_arbitrary_in, vt_max_out,vt_min_out,n_max_out,n_min_out,flag_out)
+
+
+    # Go through the list of maxima, calling time_point_thresh (the refactored version ignores the nan padding)
+    walk_forward = np.array([0],dtype=np.int32)
```
mmatteo, line 89: NaN

mmatteo (changes requested):
@SamuelBorden, thanks a lot for implementing this. It will be very useful. I made some comments to align your processors to our conventions.

mmatteo:
> Hey Sam,
> Looks good, aside from the couple of minor comments.
> Since this is our first processor that is explicitly using a variable-width vector output, I also want to double check with @mmatteo that we're all happy with the convention adopted here of preceding the vector variables with v, so a vector of time variables becomes `vt_...`, and preceding the variables that represent the size of the vectors with `n_...`.
> Ian

Looks good!

SamuelBorden:
Somehow I think I have gotten this pull request messed up because I didn't properly set the upsteam on my local branch -- sorry for that.  Should I just re-open a new pull request?

iguinn:
I think it should be ok, it looks like things merged successfully...I'd try pulling in the legend-exp dev branch, though, to make sure there aren't any weird merge conflicts.

pygama/dsp/_processors/get_multi_local_extrema.py
```diff
@@ -0,0 +1,98 @@
+import numpy as np
+from numba import guvectorize
+from pygama.dsp.errors import DSPFatal
+
+
+
+@guvectorize(["void(float32[:], float32, float32[:], float32[:], float32[:], float32[:], float32[:], float32[:])",
+              "void(float64[:], float64, float64[:], float64[:], float64[:],float64[:], float64[:], float32[:])"],
```
mmatteo, line 8: float32 -> float64 for the last variable

pygama/dsp/_processors/get_multi_local_extrema.py
```diff
@@ -0,0 +1,98 @@
...
+             "(n),(),(m) -> (m),(m),(),(),()", nopython=True, cache=True)
+def get_multi_local_extrema(w_in, delta_in, t_arbitrary_in, vt_max_out, vt_min_out, n_max_out, n_min_out, flag_out):
```
mmatteo, line 10: is delta an amplitude? If so I would use a_delta_in

We need to define a convention for a variable such as flag_out. @iguinn @ggmarshall, any suggestion?

mmatteo (changes requested):
Hi @SamuelBorden, it looks great. 

I have just a couple of trivial comments. In addition, I'd like to set a convention for variables such as flag_out. This is of general interest

jasondet:

PR #170, opened 2021-07-20
merge chloe's and jason's Ecal improvements

jasondet:


jasondet:
@mmatteo or @wisecg let me know if you want to review these before I merge them

wisecg:
@jasondet @cransom12 Nice work here.   There are so many changes here that I recommend someone upload an example script which uses the functions in analysis/calibration.py, etc.

wisecg (commented):
@jasondet @cransom12 I recommend we add an example script.  There are so many changes & new functions that it's difficult to see what the recommended workflow should be now.

mmatteo (approved):
I know that @Kermaidy @ggmarshall and @abialexander have tested the new code. They might have received some instructions from Chloe and could provide a working example as suggested by @wisecg

Kermaidy:
@mmatteo 
I've seen this thread and will check with @abialexander what we can provide.

jasondet:
Great thanks guys. In that case, I'll merge this and we'll get the example codes provided ASAP

PR #180, opened 2021-08-09
Updated dsp_optimize and added new processors

sweigart:
This commit updates dsp_optimize.py to allow for non-scalar objects to be returned from the figure-of-merit function and adds three new DSP processors.  Details on the new functionality in dsp_optimize.py can be found in the slides shown at the [Analysis Call on June 17, 2021](https://indico.legend-exp.org/event/603/contributions/3001/attachments/1703/2613/sweigart_update.pdf).

### DIFF
--- a/pygama/dsp/ProcessingChain.py
+++ b/pygama/dsp/ProcessingChain.py
@@ -461,7 +461,7 @@ class ProcessingChain:
                 if isinstance(var, np.ndarray) and len(var.shape)==2:
                     return var.shape[1]
                 else:
-                    raise ProcessingChainError("len(): " + node.args[0].id + "has wrong number of dims")
+                    raise ProcessingChainError("len(): " + node.args[0].id + " has wrong number of dims")
             elif func=="round" and len(node.args)==1:
                 var = self.__parse_expr(node.args[0], allocate_memory, var_name_list)
                 if var is None: return None
--- a/pygama/dsp/_processors/optimize.py
+++ b/pygama/dsp/_processors/optimize.py
@@ -1,0 +1,121 @@
+import numpy as np
+from numba import guvectorize
+from .pole_zero import pole_zero, double_pole_zero
+from iminuit import Minuit
+
+class Model:
+    """
+    The model class containing the function to minimize.
+    """
+    errordef = Minuit.LEAST_SQUARES
+
+    # the constructor
+    def __init__(self, func, wf_in, baseline, beg, end):
+        self.func = func
+        self.x    = np.arange(beg, end)
+        self.y    = np.asarray(wf_in, dtype=np.float64) - baseline
+        self.beg  = beg
+        self.end  = end
+
+    # the function to minimize
+    def __call__(self, args):
+        y_pz = self.func(self.y, *args)[self.beg:self.end]
+        return np.abs(np.sum(self.x) * np.sum(y_pz) - len(self.x) * np.sum(self.x * y_pz))
+
+@guvectorize(["void(float32[:], uint16 , int32, int32, float32, float32[:])",
+              "void(float64[:], uint16 , int64, int64, float64, float64[:])",
+              "void(float32[:], float32, int32, int32, float32, float32[:])",
+              "void(float64[:], float64, int64, int64, float64, float64[:])"],
+             "(n),(),(),(),()->()", forceobj=True)
+def optimize_1pz(wf_in, baseline, beg, end, p0, val0):
+    """
+    Find the optimal, single pole-zero cancellation's parameter
+    by minimizing the slope in the waveform's specified time range.
+    
+    Parameters
+    ----------
+    wf_in   : array-like
+              The input waveform
+    baseline: int
+              The resting baseline
+    beg     : int
+              The lower bound's index for the time range over
+              which to optimize the pole-zero cancellation
+    end     : int
+              The upper bound's index for the time range over
+              which to optimize the pole-zero cancellation
+    p0      : int
+              The initial guess of the optimal time constant
+    val0    : int
+              The output value of the best-fit time constant
+              
+    Processing Chain Example
+    ------------------------
+    "tau0": {
+        "function": "optimize_1pz",
+        "module": "pygama.dsp.processors",
+        "args": ["waveform", "baseline", "0", "20*us", "500*us", "tau0"],
+        "prereqs": ["waveform", "baseline"],
+        "unit": "us"
+    }
+    """
+    m = Minuit(Model(pole_zero, wf_in, baseline, beg, end), [p0])
+    m.print_level = -1
+    m.strategy = 1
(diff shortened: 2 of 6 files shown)


jasondet:
@iguinn can you take a look?
@mmatteo please check the saturation and soft_pilup_corr processors

mmatteo:
@sweigart,  these processors are going to be useful.

Before merging, can you please check the template.py processor and add to all your processors the steps/conventions mentioned there?

For instance:
- input/output parameters should all be float32 or float64 (in total two guvectorize declarations)
- all output parameters should be initialized to nan at the beginning of the processors
- add checks that none of the input parameters is none
- we also have a convention on the name of the variables, e.g tau -> t_tau

sweigart:
@mmatteo, thank you for reviewing the new processors.

I’ve reviewed template.py and conformed the new processors to the conventions laid out there.  This includes implementing the suggestions you listed.  However, I didn’t change your example of `tau` to `t_tau` per
https://github.com/legend-exp/pygama/blob/b0f52f772a7f390b5ebd18a68e6a625edf6d7949/pygama/dsp/_processors/template.py#L26
because `tau` is a time constant, not an index.

These conventions, though, don't seem to be established yet.  I could only find one processor—fixed_time_pickoff.py—that conformed to all of the conventions mentioned in template.py.  As but one example, `bl_subtract` violates your first and last bullet points:
https://github.com/legend-exp/pygama/blob/b0f52f772a7f390b5ebd18a68e6a625edf6d7949/pygama/dsp/_processors/bl_subtract.py#L5-L9
where `uint16` should be `float32` or `float64` and `a_baseline` should be `a_baseline_in`.

In general, the conventions laid out in template.py seem reasonable to me except for the `w_`/`t_`/`a_` prefixes since I don't understand the motivation behind them.  These three prefixes are somewhat specific to our use case for these processors and don't cover the full range of possible function arguments, e.g., the aforementioned `tau`.

I can see a general use for such prefixes, though, to make some assumptions clearer in the code.  For example, an `i_` prefix could be used on variables that are assumed to be integer values despite being `float32` or `float64` (due to the NaN convention).  As another example, there could be a prefix to denote whether an output variable is meant to be an array or a scalar (since this is obscured by using `[:]` for all output variables in the Numba decorator).  That said, having all of these prefixes and suffixes could defeat the main purpose of adding clarity by making the code cumbersome to read.  Alternatively, we could use just a `i_`/`o_` prefix to replace the `_in`/`_out` suffix, which I find easier to read.

If desired, I can conform the existing processors to these conventions in a separate pull request and, at the same time, propose a hopefully clearer convention for the prefixes/suffixes.

ggmarshall:
@sweigart I agree it makes more sense for tau to be by itself, looking through the processors I think it is more in keeping for all constants to have stand alone names. The point of the w_/t_/a_ is then to identify arguments that are waveform dependent. Maybe a better convention would be to have the argument type (w_/t_/a_) followed by whether it is input or output and then followed by an optional descriptor. So the baseline would now be a_in_baseline. What's everyone's thoughts on this?

Regarding uint16, I found that there were difficulties casting the uints into floats as such an exception to the float convention was needed for the baseline.

sweigart:
@ggmarshall, thank you for your feedback.

In regard to the `w_`/`t_`/`a_` prefixes, I now understand that you wanted to flag which variables concern waveforms.  It wasn't entirely clear to me since there are cases such as: https://github.com/legend-exp/pygama/blob/b0f52f772a7f390b5ebd18a68e6a625edf6d7949/pygama/dsp/_processors/pole_zero.py#L9 where a time constant has a `t_` prefix.  There are also cases where output amplitudes don't have the `a_` prefix (or an `_out` suffix) that I'd expected to see.  For example: https://github.com/legend-exp/pygama/blob/b0f52f772a7f390b5ebd18a68e6a625edf6d7949/pygama/dsp/_processors/linear_slope_fit.py#L9

The above example from linear_slope_fit.py also serves as a good example of why I'd said that these prefixes seem a bit too specific.  The `slope` output depends on the waveform, but, strictly speaking, it's not an amplitude—so what prefix would be used in that case?  Do we need to generalize the prefixes' meanings or define more of them?

More importantly, I'm still unclear on _why_ one needs to know that a given variable depends on a waveform—and, in part, how they depend via the different prefixes—while not others.  I feel that such information should already be clear by having sufficient documentation describing the parameters at the beginning of the function (as per template.py).

My vote is to just use an `i_`/`o_` prefix (and no suffix) on _all_ function arguments to denote whether it is an input or output.  While it can be argued that even this is unnecessary, I do think that it adds a bit of clarity when reading the code, allowing for a quick distinction between them and internal variables.

Adding many prefixes and/or suffixes could risk making the code cumbersome and, in turn, harder to read.  I also think that the meaning of the `i_`/`o_` prefixes would be intuitive to folks after looking at the code a bit—even if they didn't read template.py beforehand.  In contrast, I would probably intuit the meaning of `w_` but not `t_` or `a_` without seeing template.py first.  Further, if it's useful to know that a given variable depends on the waveforms somehow, it should be clear from the variable name itself, e.g., as it already is to me for the `mean`, `stdev`, `slope`, and `intercept` variables in linear_slope_fit.py above.

In regard to the `int` usage, there are also other processors that don't conform to: https://github.com/legend-exp/pygama/blob/b0f52f772a7f390b5ebd18a68e6a625edf6d7949/pygama/dsp/_processors/template.py#L14

As a couple of additional examples: https://github.com/legend-exp/pygama/blob/b0f52f772a7f390b5ebd18a68e6a625edf6d7949/pygama/dsp/_processors/time_point_thresh.py#L6-L10 https://github.com/legend-exp/pygama/blob/b0f52f772a7f390b5ebd18a68e6a625edf6d7949/pygama/dsp/_processors/presum.py#L3-L8

In regard to `uint16` specifically, the DAQ baseline was also the reason why I'd used it in optimize.py.  I tried running the processor using only floats yesterday, though, and found that it handled the conversion to `float32` successfully.  In any case, such an issue has already been solved in build_processing_chain.py for the DAQ waveform, which gets converted automatically from `uint16` to `float32` on this line: https://github.com/legend-exp/pygama/blob/b0f52f772a7f390b5ebd18a68e6a625edf6d7949/pygama/dsp/build_processing_chain.py#L160

We could do something similar for the baseline.  That would remove the need for such exceptions.  If needed, I'm willing to add that in when conforming all of the processors to the desired conventions.

sweigart:
I've submitted pull request https://github.com/legend-exp/pygama/pull/183, by the way, which conforms all of the existing processors to the conventions mentioned in template.py.  I didn't change the function argument names, though, because there doesn't appear to me to be a consensus yet on what convention to use—if any.

Perhaps we should continue the discussion on conventions there (or in a new issue).  When the conventions to be used are settled, I'll update the new processors in this pull request as appropriate.

jasondet:

PR #181, opened 2021-08-10
Added LAr SiPM processing

SamuelBorden:
Added "processing.py" in which users can specify the files and the channels they want to process. The script follows the format created by David Sweigart for multi-threading on the channels you want to analyze. The current three processors in this file output the maximum, the time of the maximum, and the total integral of each waveform.


jasondet:
@iguinn and @mmatteo can you take a look at these processors?
@gipert can you also take a look at the code in lar-commissioning?

pygama/dsp/_processors/bl_estimator.py
```diff
@@ -0,0 +1,29 @@
...
+           wf to baseline subtract
+    w_out : array-like
+            The output waveform from the processor
+    """
+    
+    w_out[:] = np.nan
+    if (np.isnan(w_in).any()):
+        return
+    if (len(w_in)<250):
+        raise Exception('window of averaging is out of range')
```
iguinn, line 26: Should use DSPFatal from pygama.dsp.errors instead of Exception

pygama/dsp/_processors/bl_estimator.py
```diff
@@ -0,0 +1,29 @@
...
+        
+    w_out[:] = w_in[:] - (w_in[0:250].mean()+w_in[-250:].mean())/2
```
iguinn, line 28: Do we want 250 samples to be hard-coded here, or should we include this as an input parameter?

jasondet:
As discussed with Sam and Luigi, we are going to close this PR and accept Luigi's as the starting point (after Luigi commits a few missing pieces)

PR #183, opened 2021-08-19
Processor cleanup

sweigart:
These changes follow up on the discussion in pull request https://github.com/legend-exp/pygama/pull/180.

A set of conventions for the DSP processors—described in template.py—was meant to be established.  This pull request strives to conform the current processors to those desired conventions.  It also strives to format the code in a uniform manner so as to make it easier to read.  The only convention from template.py not established here is the prefixes/suffixes in the function argument names because that discussion seems to be still ongoing.

I've tested these changes on the PGT calibration data using some of the analysis scripts on the `exp-lpgta` branch.  As far as I can tell, the results are the same before and after these commits.  However, my scripts do not utilize all of the DSP processors that have been updated.  It'd be helpful if someone could review and/or test a different subset of processors used in their analysis procedure.  Perhaps after the major refactor, we should develop a unit test for each processor so that such changes in the future can be easily checked.

### DIFF
--- a/pygama/dsp/_processors/bl_subtract.py
+++ b/pygama/dsp/_processors/bl_subtract.py
@@ -2,39 +2,35 @@ import numpy as np
 from numba import guvectorize
 from pygama.dsp.errors import DSPFatal
 
-@guvectorize(["void(float32[:], uint16, float32[:])",
-              "void(float64[:], uint16, float64[:])"],
+@guvectorize(["void(float32[:], float32, float32[:])",
+              "void(float64[:], float64, float64[:])"],
              "(n),()->(n)", nopython=True, cache=True)
-
 def bl_subtract(w_in, a_baseline, w_out):
     """
-    Processor to subtract the fpga baseline from all waveform values. If any
-    input values are nan will return array of nan of w_out.
+    Subtract the constant baseline from the entire waveform.
+
     Parameters
     ----------
     w_in : array-like
-           wf to baseline subtract
-    a_in : uint
-           The baseline values to subtract, these are stored as uints in the raw files
-    w_out : array-like
-            The output waveform from the processor
+           The input waveform
+    a_in : float
+           The baseline value to subtract
+    w_out: array-like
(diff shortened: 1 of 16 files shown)


ggmarshall:
Hi David, looks good to me, I only have a couple of comments:
1) Regarding bl_subtract if we cast the baseline from uint to float in build processing chain as suggested in #180 then this processor can just be removed as we can use numpy subtract instead.
2) I'm not sure about the np.floor checks. I also originally had these in the processors but removed them as it meant that every time value in the dsp config had to be a multiple of the sampling rate. Maybe this should instead be handled in the processing chain so that the dsp config is independent of the sampling rate.

sweigart:
@ggmarshall, thank you for looking over these changes!

1.  In normal processing, I agree that the bl_subtract processor could be replaced with NumPy's subtract.  However, when, say, one of the values of `w_in` is `np.nan`, then bl_subtract would make all values in `w_out` be `np.nan`s while NumPy's subtract would only cause one of the values in `w_out` to be `np.nan`.  Here's an example:
```
>>> import numpy as np
>>> x = np.arange(5, dtype='float')
>>> x[1] = np.nan
>>> np.subtract(x, 2)
array([-2., nan,  0.,  1.,  2.])
```
So, there would be a substantive difference in using bl_subtract versus np.subtract.  From template.py (and some of the other processors), my impression was that we'd want the entire output waveform to be `np.nan`s.  My recommendation would be to keep bl_subtract for now until that we decide on what precise functionality we want (because it doesn't interfere with anything else in pygama by being there).

2.  I also noticed that consequence of the np.floor checks, and I had to update the DSP configurations on the `exp-lpgta` branch to use exact multiples of the sample rate (as you said).  One advantage of keeping the np.floor checks is that it forces us to know exactly what number of samples we are (and want) to use.  In any case, I added them because that's the (desired) convention that is currently described in template.py.  So, if we don't want them after all, we should update template.py to reflect that—which can be done in a separate pull request.

jasondet:

PR #186, opened 2021-09-01
Added multi-a processor

SamuelBorden:
This processor calls the get_multi_local_extrema function and returns the amplitudes of maxima found in a waveform


jasondet:
I was told that this already matches the conventions discussed last week. I'm going to go ahead and merge it, and we can edit it further as necessary.

PR #187, opened 2021-09-06
Scripts for processing LAr commissioning data

gipert:
Here are couple of basic scripts I initially wrote to run DSP on SiPM data and a PE spectrum of PGT run 28 I produced with them:

image


gipert:
I added some missing scripts and improved the docs, hope it is usable now. Give it a look and let me know if you think we can build on this. The DSP configs are super preliminary, I expect them to be updated soon.

CC @rosannadeckert @SamuelBorden @nickruof

PR #188, opened 2021-09-06
Remove Python v3.[5-6] from testing

### LINKED ISSUE #178: Minimum supported Python version
I realized that CI builds with Python versions 3.5 and 3.6 fail because of the `h5py>=3.2.0` constraint being unsatisfiable. Should we now explicitly support Python >= 3.7?

gipert:
Fixes #178. I've tried adding Python v3.9 but looks like pyFFTW is not ready yet: https://github.com/pyFFTW/pyFFTW/issues/301


jasondet:
very nice, thanks Luigi!

PR #190, opened 2021-09-14
dev

gipert:
- Added one dimensional Gaussian filter for dsp based on scipy's gaussian_filter1d function
- Reordered factory function so that kernel isn't recomputed every function call. Factory function generates and returns the guvectorized function that is the convolution between the kernel and the signal. Also added error message if the kernel is much larger than the signal
- implement poly_match and get_i_local_extrema
- Add routine for extracting a simple scaling factor
- Add poly function and improve documentation
- get_hist makes copy of hist for var, add better_int_binning, and fwfm guards against dy=0 in bins_over_f method
- first pass for refactor. only poly_match and find_E_peaks tested so far.
- add input handling
- debug several Ecal function
- fix gof code and optionally use it to inflate errors
- add a useful comment
- add option to inflate peak top fit errors with chi2red
- Added radford_fwhm to peak_fitting.py
- edit poly_match for len(xx) < len(yy)
- hpge_find_E_peaks returns when poly_match doesn't match all requested peaks
- fix bug in find_bin
- include missing dependency
- fix misnamed variables
- fix bug (get_fwfm returns value + uncertainty)
- fix loop over int to range(int)
- fix slice over nparray instead of list
- Added pygama.dsp.errors. Throw DSPFatal in trap_filter and trap_norm if trap is too long. Use ProcessingChainError in ProcessingChain.
- Made type signatures less picky. If types given for a processor can be safely cast to the right types, accept the args
- ProcessingChain get_variable now has the option to parse variable names from string without allocating memory. Use this to automatically figure out prereqs instead of requiring an entry in build_processing_chain
- Clarify errors when parsing expressions with ProcessingChain
- Improved error handling in build_processing_chain
- Added windower processor
- Remove py2.7 test from dev branch
- unify return type for get_fwfm
- fix poly_match
- ensure minimum bin width of 1 for better_int_binning
- fix get_fwfm value fraction
- prevent div0 in get_fwhm
- try high stats get_fwhm, then low stats
- hpge_fit_E_peaks saves binwidth and range to results
- rebin to 0.2 keV after first step update min/max, guess_keV
- rerun peakfinder after rebinning
- rename detected/fitted peak variable names, include keV values
- catch hpge_fit_E_peak_tops and hpge_fit_E_peaks fitting failures
- remove failed fits
- Refactor pole zero
- Refactor fixed time pickoff
- Refactor linear_slope_fit remove linear_fit and mean_stdev
- Refactor log_check
- Refactor min_max
- Refactor tpt, remove time point frac
- Added bl_subtract
- Added template processor
- Merged convolutions into single file and refactored
- Merged and refactored trap filters
- Added moving windows and refactored current
- add error message for negative fwfm
- Store.read_object will now read a list of file names from any iterable object
- Requre h5py version > 3.2.0 to fix bug in lh5.Store
- Asym trap bug fix
- error handling for get_fwfm
- error handling for gauss_mode_max gauss_mode_width_max
- auto estimate Etol_keV when not given in find_E_peaks
- error handling in get_hpge_E_peak_par_guess
- hpge_fit_E_peaks now takes tuples for left/right peak ranges
- second peak_finder replaced with peak_getter
- simplify verbose output
- Added radford_fwhm to peak_fitting.py
- deal with covariance non-calculation errors
- can pass in different functional forms for peak fitting
- raise errors when get_fwfm fails
- skeleton function of get_hpge_peak_bounds
- fix bounds in radford_fwhm
- add wrapped version of radford_peak function
- add get_fwhm_func (fwhm from function and parameters)
- remove unused dependencies
- error handling
- get_hpge_E_peak_par_guess for radford_peak_wrapped
- implement get_hpge_E_peak_bounds
- sometimes covariance estimation fails differently
- actual fix this time
- Added flat int for ranges
- fix post peak fitting checks
- each energy peak can only match to one histogram peak
- correct amplitude estimation for radford_peak
- fix radford_peak bound checks
- code pruning
- make default poly_match energy tolerance smaller
- peaks_keV doesn't have to be sorted, just array
- solve deg > 1 limits correctly
- can set peaks to be located but not fitted (poly_match)
- get_fwfm in try loop
- get_fwfm mask 0 variance bins
- refactor radford_peak (don't implement tail/step twice)
- Fixed bug in Gretina4M decoder
- add get_mu_func wrapper for flexibility in parameter definition
- calculate fwhm uncertainty
- par_guess/bounds for radford_peak (unwrapped)
- peak rejected when var_mu = 0
- gauss_tail now uses an approximation in case of overflows
- remove unnecessary output
- covariance matrix excludes amp and step height
- Removed last line
- Updated to use new processors
- removed floor condition
- Added multi-t-filter
- recover accidental deletion during conflict handling
- return tail=0 immediately for zero amp, plus some formatting
- remove deletions that would give grad2 the wrong length
- Changed multi_t_filter to object mode
- Split get_multi_local_extrema into its own function
- fixes to FWHM error extraction
- Updated function description
- Fixed bug in store.py when a h5py._hl.files.File object is passed into the read_object function.
- Removed the checks of negative input parameters in the pole-zero cancellation functions.
- Removed compatibility checks for python 3.5 and 3.6
- Updated dsp_optimize to allow for non-scalar objects to be returned from the figure-of-merit function.  Added new DSP processors.
- Fix in store.read_object that speeds up the reading when an idx index selection is passed in.
- Fixed checks and initalizations in multi_t
- Conformed new processors to the conventions in template.py.
- First pass of conforming the processors to the conventions in template.py.  Cleaned up syntax and made it consistent.
- Minor fixes to processors after testing with PGT data.
- Fixed function signature + a_delta_in
- Fixed resize bug in Array
- Revert "Fixed resize bug in Array"
- build_processing_chain now returns a field_mask
- added multi-a processor
- changed variable names to follow template
- implement channel-specific wf lengths
- Add dependency iminuit in setup.py


gipert:
Sorry, opened by mistake!

PR #193, opened 2021-09-15
Force numba version <= 0.52.0

iguinn:
See https://github.com/numba/numba/issues/7402


oschulz:
Looks like this will be fixed in numba v0.55: https://github.com/numba/numba/pull/7403

PR #199, opened 2021-10-06
Nruof/tqdm bar

nickruof:
This replaces the current progress bar, used in raw_to_dsp, and uses the tqdm python package progress bar.  For a cleaner look and also includes iteration times and finish time estimates.  

There are many ways to customize the bar, but left it looking as close to what is already there, with the added functionality of changing the length and changing the text that is displayed in front of the bar.

Screenshot from 2021-10-05 19-01-24

### DIFF
--- a/pygama/utils.py
+++ b/pygama/utils.py
@@ -54,7 +54,7 @@ def sh(cmd, sh=False):
     return decoders
 
 
-def tqdm_range(start, stop, step=1, verbose=0, text="Processing", bar_length=20):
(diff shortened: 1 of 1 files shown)


wisecg:
Cool!  I use tqdm all the time in my own code, it's the best.

Fully support merging this.

gipert:
This breaks parts of pygama that still use the old method, see for example `io/fcdaq.py`. The new progress bar should be used consistently in the code base. @nickruof can you fix this asap in a new PR, please?

nickruof:
Addressed in #203

PR #203, opened 2021-11-02
Update all progress bars to tqdm

nickruof:


gipert:
Hi Nick, I would be really happy if pygama used only one progress bar (`tdqm`), we should try to keep the codebase clean. The built-in progress bar is also buggy. Have you got a bit of time to work on this? In practice, there are three loops in `fcdaq.py`, `llamadaq.py` and `orcadaq.py` that should be updated to use `tdqm`. I can then test the changes and make sure that nothing is broken.

nickruof:
Ok I will update all progress bars to use the tqdm one and add to this PR

nickruof:
Here's what it looks like for flashcam data
Screen Shot 2021-11-05 at 11 32 03 AM

wisecg:
This looks good to me.  In prep for the analysis workshop I'm going to go ahead and merge this so that it's included on the master branch.

PR #211, opened 2021-11-22
add updated 3316 readout code

wisecg:
The COHERENT HPGe working group has been using pygama to read out their Struck 3316, and James Browning (@jbrowni2) was able to update the old 3316 code in pygama with a format much closer to "new" pygama conventions -- using LH5, similar to the Struck 3302.

This may not have all the features we would want in the final version, but it works!


wisecg:
Whoops, this added a tutorial notebook that should have gone in a different PR to the dev branch.  Merging this one, and will follow up with the rest of the notebooks in a separate PR.

PR #220, opened 2021-12-06
support separate DSP configs for individual channels

wisecg:
I needed a way to process two channels in the same LH5 file with different DSP configs.   I didn't want to call `raw_to_dsp` twice, because then I have to be careful about overwriting the existing file when processing the second channel.    Instead, I added a new option, `chan_config` which if set, gives a lookup table between the group to process, and the location of the JSON DSP file.  Example from the CAGE code:
```
lh5_tables = ['ch146/raw', 'ch150/raw']
chan_config = {'ch146/raw':f_config,
               'ch150/raw':f'{dsp_dir}/dsp/dsp_nplus.json'}

raw_to_dsp(f_raw, f_dsp, f_config, n_max=nwfs, verbose=verbose,
                   overwrite=overwrite, lh5_tables=lh5_tables, chan_config=chan_config)
```
Since it's just a new keyword argument, I don't think this will break anybody's existing code.

I also changed the location of the `dsp_info` table, so that it would still be able to store the separate `dsp_config` dictionaries used to process each file.  

```
18:56 wisecg@cori04:processing$ h5ls /global/cfs/cdirs/m2676/data/cage/LH5//dsp/cage_run385_cyc2527_dsp.lh5/dsp_info
dsp_config               Group
h5py_version             Dataset {SCALAR}
hdf5_version             Dataset {SCALAR}
numpy_version            Dataset {SCALAR}
pygama_branch            Dataset {SCALAR}
pygama_date              Dataset {SCALAR}
pygama_revision          Dataset {SCALAR}
pygama_version           Dataset {SCALAR}
python_version           Dataset {SCALAR}
timestamp                Dataset {SCALAR}
18:56 wisecg@cori04:processing$ h5ls /global/cfs/cdirs/m2676/data/cage/LH5//dsp/cage_run385_cyc2527_dsp.lh5/dsp_info/dsp_config
ch146                    Group
ch150                    Group
```

 If `chan_config` isn't set (which is the default for everybody else's code right now), then `dsp_config` will appear as it did before.

@jasondet @iguinn Any objections to merging this?   It's pretty flexible and shouldn't break any existing code.

### DIFF
--- a/pygama/io/raw_to_dsp.py
+++ b/pygama/io/raw_to_dsp.py
@@ -21,19 +21,12 @@ from pygama.dsp.errors import DSPFatal
 
 def raw_to_dsp(f_raw, f_dsp, dsp_config, lh5_tables=None, database=None,
                outputs=None, n_max=np.inf, overwrite=True, buffer_len=3200,
-               block_width=16, verbose=1):
+               block_width=16, verbose=1, chan_config=None):
     """
     Uses the ProcessingChain class.
     The list of processors is specifed via a JSON file.
     """
     t_start = time.time()
-
-    if isinstance(dsp_config, str):
-        with open(dsp_config, 'r') as config_file:
-            dsp_config = json.load(config_file, object_pairs_hook=OrderedDict)
-
-    if not isinstance(dsp_config, dict):
-        raise Exception('Error, dsp_config must be an dict')
 
     raw_store = lh5.Store()
     lh5_file = raw_store.gimme_file(f_raw, 'r')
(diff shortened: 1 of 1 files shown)


iguinn:
Would it be make sense to combine the chan_config and dsp_config into a single argument, and have it act as before if it's a string, and separate things out if it's a dict or a list, or something like that?

wisecg:
I thought about that exact approach - putting in a test for dsp_config being a list.  The reason I didn't do it is that I needed to be able to do a dict lookup with my list of lh5_tables that i'm putting in, so I would still have had to supply another dictionary with the matching.  You can't assume a 1-1 matching between lh5_tables and dsp_config in a list, because of the way the code gets rid of lh5_tables it doesn't recognize.

iguinn:
Oh I see, in that case I'm fine with merging

PR #221, opened 2021-12-12
Add hit-to-event (hit_to_evt) processing step

wisecg:
This follows the structure of the `daq_to_raw` and `raw_to_dsp` apps already existing in pygama/io.

The main workhorse is `cluster_events`, which does most of the event building work.   For simplicity, we don't consider the case of building across multiple cycle files, though in principle it could be added.   `cluster_events` takes an input list of DataFrames, and uses some nice pandas magic to create the event-built columns, and sort all the input data by strictly ascending timestamps.   I think that this "sort_values" step should not be implemented strictly in terms of LH5 objects, because 1) a copy is inevitable because we have to permute the order of rows in the input tables to create the output event-built table, and 2) the pandas functions made creating the time-coincidence map very easy, especially the `groupby` I used to create the "sub-event" column.

@iguinn @jasondet Comments welcome!

### DIFF
--- a/pygama/io/hit_to_evt.py
+++ b/pygama/io/hit_to_evt.py
@@ -1,0 +1,239 @@
+#!/usr/bin/env python3
+import os
+import time
+import argparse
+import h5py
+import numpy as np
+import pandas as pd
+from pprint import pprint
+
+import pygama.lh5 as lh5
+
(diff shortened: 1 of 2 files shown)


wisecg:
I'm taking silence as tacit approval 😄 so I'm going to go ahead and merge this into dev.

PR #227, opened 2022-02-02
Upgrade to numba>=0.55

iguinn:


wisecg:
I think we should try testing py3.10 and the new numba on the dev branch first, and process some data or something first, before merging into master.

gipert:
@iguinn we can use this PR to already check whether the pip-install step works with newer Python versions on GitHub Actions. Can you update `.github/workflows/main.yml` and push a commit here?

iguinn:
Ok, I added it but it seems to be testing 3.1 instead of 3.10...

iguinn:
Ok, fixed

wisecg:
Considering that the numba version doesn't work with earlier versions, I propose we drop the other python version tests @gipert @iguinn and only include e.g. 3.10

gipert:
Well, now @iguinn allowed for `numba<0.53` too, so older Python versions might be still used (with a compatible Numba version). Right?

jasondet:
Ian says this can be merged at this point.

PR #229, opened 2022-02-03
Strip metadata from jupyter notebooks

gipert:
The Git filter that cleans up Jupyter notebooks on my laptop keeps modifying the upstream files. I cannot ignore it since it takes forever to run any Git command and I always have these unstaged changes around. The modifications make sense to me.

This commit should be propagated to `dev` and possibly other branches.

### DIFF
--- a/tutorials/DataProcessing.ipynb
+++ b/tutorials/DataProcessing.ipynb
@@ -154,22 +154,8 @@
   }
  ],
  "metadata": {
-  "kernelspec": {
-   "display_name": "legend-base",
-   "language": "python",
-   "name": "legend-base"
-  },
   "language_info": {
-   "codemirror_mode": {
-    "name": "ipython",
-    "version": 3
-   },
-   "file_extension": ".py",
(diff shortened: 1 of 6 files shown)


iguinn:
Hey Luigi, thanks for bringing this up. The filter is added during `setup.py` at line 94, and right now just clears the output. It can also apparently clear metadata by adding `--ClearMetadataPreprocessor.enabled=True`, which seems like it's a good idea. Can you try adding it and see if it helps things? The goal of the filter was to make it so we could add jupyter notebook tutorials without creating a lot of merge conflicts and such, but if you or anyone else has other ideas of how to handle this, we should consider that as well.

gipert:
Ok now I understand what's happening. I have a Jupyter version which is recent enough to get the additional `--ClearMetadataPreprocessor.enabled=True` flag. This is why the filter strips the metadata too on my system. I assume the last one who used the notebooks and committed had an older Jupyter version. I would go ahead and merge, if no objections.

PR #231, opened 2022-02-15
Minor fixes for sipms processors to work in processing chain

ggmarshall:
Small changes to signatures to make these work in processing chain, removed the arbitrary input arrays and removed the call to get local extrema instead run as a separate processor before the multi_t and multi_a filters.  Also added all these to the processors.py file.


jasondet:
@iguinn and/or @SamuelBorden these look good to me, can one of you check them too?

SamuelBorden:
These look great, thanks @ggmarshall !

PR #232, opened 2022-02-15
Enabling tests

gipert:
- [x] Enable GitHub action tests on OSX
- [x] Use pytest rather than unittest, `pip install` it in CI jobs
- [ ] Test sub-folder per module, with micro tests for each function. Should test on syntetic traces rather than real?
- [ ] High-level testing of the full production chain on real data:
  - Fetch `daq` test files from <https://github.com/legend-exp/legend-testdata>
    (note: the repo is and should stay public)
  - What should it be? A couple of (coincident) FC files from HPGe + SiPM streams
    to really be a 100% realistic?
  - I guess we need test files from other data-takers?
- [ ] What else would be interesting to test?
- [ ] Test coverage?

### DIFF
--- a/.github/workflows/get-dependencies.sh
+++ b/.github/workflows/get-dependencies.sh
@@ -9,4 +9,4 @@ else
     sudo ln -s /usr/bin/llvm-config-8 /usr/bin/llvm-config
 fi
 
-python -m pip install --upgrade pip wheel setuptools numpy GitPython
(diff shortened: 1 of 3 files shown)


gipert:
Let's merge this for now. @jasondet can you have a quick look and push that button?

jasondet:
Thanks Luigi!

Issue #235, opened 2022-02-28
Passing user-created data into processing chain


SamuelBorden:
For some processors we want to load in specific data that the user has created. For example, for a Wiener filter we want to pass a superpulse that the user has made using whatever superpulse building code they want. I have already talked with @iguinn and he has identified at least two ways to do this:

1. >  Put a json-array of the values into the DB file. I'm not 100% sure this is supported yet but it's something we will want.
2. > Pass the factory the name of a file to read the waveform from.

> Which one makes more sense probably depends a little on how the metadata processing framework looks [...] Approach 1 is more direct, but would make the json DB files a lot bigger and less readable.
>  Approach 2 could also have other advantages, since you'd be able to have your super-pulse building code write it however you want, and then have the processor read it that same way.

I'm interested in hearing others' opinions on this, especially with regard to the metadata framework we have.

mmatteo:
how would you parse info such as wf length, sampling period and time offset in option 2?

wisecg:
I like Approach 2 a lot more.  Keep the DSP files human-readable and relatively small, and just specify auxiliary information to be read in by filename

@mmatteo One way Sam can do this is as follows -- Each function called by processing chain has an "initialize" step where we can read in values from a database, etc, before the main processing loop begins.   Each function which needs to read in a "library file"  from a filename will have specific procedures to load the data into whatever format is needed for a computation.

iguinn:
Yeah, I agree with Clint I think. To get more granular, what it would look like, is Sam would write a factory-function that takes as an input a filename, reads in an array (and this could in theory have whatever format he wants, but it may make sense for us to come up with a standardized way to do this), and then guvectorize compiles a function around it that takes a waveform as an input and spits out the deconvolved one as an output. We don't have anything directly analogous, but for examples, look at `fftw.py` (since you're going to be using ffts for this) and perhaps `gaussian_filter_1d.py`, which generates a kernel from input parameters (instead we would generate a kernel from an input filename)

jasondet:
Looks like the Wiener filter solution will work, marking this closed.

PR #238, opened 2022-03-07
Added first pass of Wiener filter

SamuelBorden:
Following the discussion in issue #235 this factory function takes an lh5 file as an input to compute the Wiener filter and apply it in the frequency domain


jasondet:
Ian says it looks good.

PR #244, opened 2022-03-26
Restructuring the existing documentation

gipert:
In 560e4449519e0fb89c6593e83a3964ed6576b3c1 I migrated all the remaining docs in READMEs to the source code. I also added some missing `__init__.py` files here and there. @jasondet you can merge this anytime, if you've got no objections.

I would like to discuss all this at the next pygama dev call and recruit volunteers.

Reference issue: #240

Issue #241, opened 2022-03-24
Unit testing


gipert:
Here's a to-do list, I'll kep it up-to-date. Most of this will be implemented on the `refactor` branch only.

- [ ] Add micro-tests for each module/function separately
- [ ] Test each processor separately by using a pulse injector
- [ ] Add high-level tests of the full production chain by using real data from https://github.com/legend-exp/legend-testdata
- [ ] Test in new LEGEND Ubuntu container
- [ ] Let a Python linter run as part of tests to enforce uniform coding conventions
- [ ] Add test coverage reporting through https://codecov.io and GitHub bot
- [ ] Add Dependabot (https://github.com/dependabot) to keep Python dependencies up to date?

jasondet:
test installation doesn't work for me:
```
pygama % pip install .[test]
zsh: no matches found: .[test]
```
Also just running pytests fails:
```
pygama % pytest
ERROR: usage: pytest [options] [file_or_dir] [file_or_dir] [...]
pytest: error: unrecognized arguments: --cov=pygama
  inifile: /Users/jasondet/Work/Coding/pygama/pyproject.toml
  rootdir: /Users/jasondet/Work/Coding/pygama
```
@gipert can you take a look?

gipert:
This is zsh... you need to wrap `[]` in single quotes, like this: `pip install '.[tests]'`.

I've forgotten zsh is now the default shell on OSX.

jasondet:
ah that works! can you update the instructions?

incidentally, using the single quotes also works in bash... maybe make that the instruction?

also, the instructions should specify where to start from (i.e. `.'[test]'` inside the pygama dir, `pygama.'[test]'` from the dir containing pygama)

slwatkins:
What is the plan for unit tests and code coverage for numba-ized functions? (e.g. the processors) We can write unit tests, but because of the JIT compilation, the actual code itself is not counted as covered (see https://github.com/numba/numba/issues/4268).

Should we exclude these functions from the overall coverage? We should still write tests for them of course, but it doesn't seem like we would ever be able to get pytest to mark these blocks of code as "covered".

gipert:
Good point @slwatkins. What about disabling JIT in unit tests? Would that fix the problem?

slwatkins:
[...]

PR #245, opened 2022-03-29
Unit tests cleanup

gipert:
Reorganizing and fixing the existing tests. **Merge after #244 (if tests are passing!)**

Reference issue: #241


gipert:
Tests are failing because of the missing `__init__.py`, they should be fixed after #244 is merged.

gipert:
Uhm I don't really expect these failures (also why not in the container?). @jasondet do you understand more than I do?

gipert:
Got it, now tests are finally passing. There are some warnings though:

https://github.com/legend-exp/pygama/runs/5758376762?check_suite_focus=true#step:5:18

I think we should be more strict and make tests fail even on warnings. I will add this and fix them.

gipert:
Container builds will be fixed once we patch the Ubuntu container: https://github.com/legend-exp/legendexp_legend-base_img/pull/5

I will configure the GitHub actions to stop testing on the CentOS container.

gipert:
The LEGEND container is now too big to be downloaded on GitHub Actions runners: https://github.com/legend-exp/pygama/runs/5858421030?check_suite_focus=true#step:2:237

I will temporarily disable testing in containers until we find a solution. I'll go ahead and merge this PR.

PR #248, opened 2022-04-01
Refactor

iguinn:
- Resolved merge conflicts. build_dsp works
- Added some missing __init__ files
- Moved metadata directory of json files from tutorials to the project tree. These are now installed and can be accessed using `with importlibs.resources.path("pygama.metadata", "file.json") as p: ...`
  - Tutorials need to be updated still
- Added keyword arguments for "args" lists. Also made some fixes so that None and string values can be passed
- Some processors have time values passed as floats that are then checked for integer values. This causes problems for the refactor branch, since unit conversions often introduce roundoff errors, and unitted numbers cannot be rounded easily. I will create an issue for this.


gipert:
We should merge this after #241, there are some conflicts between the two. Just couple of comments from my side concerning packaging:

I would strongly recommend to keep user code separate from the pygama codebase, so I would not move those JSON files from `tutorials` to `pygama.metadata`. I think we should retain only code that is needed for pygama to work and keep apps/examples at least in a separate directory outside the package tree – but I would actually be much happier with a separate Git repository, something like `pygama-contrib`. We can add this as a discussion item at the next pygama call, I wanted to bring this issue up anyways.

About third-party, non-python code like Eigen: I think this should also be placed elsewhere, maybe in a folder at the repo root dedicated to third party projects. We should think about it also install-wise. I'm also usually more in favor of directly mirroring the code in our repository, instead of using Git submodules, which are sometimes painful to handle. Let's discuss this too next time we meet!

iguinn:
Thanks for the comments, Luigi. I've:
- Removed eigama; that was a mistaken inclusion, and it remains on a separate branch (as you say, there are a lot of open questions about how to include it)
- Reverted my move of metadata into the main package
- Removed the dsp README file, and expanded the __init__ documentation a little bit

As for the remaining merge conflicts, once we get #241 in, I can revisit the merge conflicts.

pygama/dsp/__init__.py
```diff
@@ -10,4 +10,6 @@
   processors for inclusion in the analysis.
 * :class:`ProcessingChain`: A class that manages and efficiently
   runs a list of DSP processors
+* :function:`build_processing_chain`: A function that builds a ProcessingChain using lh5 formatted input and output files, and a json configuration file
+* :function: `build_dsp`: A function that runs build_processing_chain to build a ProcessingChain from a json config file and then processes an input file and writes into an output file, using the lh5 file format
```
gipert, line 14: Just a small thing:
```suggestion
* :func:`build_processing_chain`: A function that builds a ProcessingChain using lh5 formatted input and output files, and a json configuration file
* :func:`build_dsp`: A function that runs build_processing_chain to build a ProcessingChain from a json config file and then processes an input file and writes into an output file, using the lh5 file format
```
Otherwise it won't be rendered correctly in the docs. Sphinx cross-referencing for Python objects is documented [here](https://www.sphinx-doc.org/en/master/usage/restructuredtext/domains.html#cross-referencing-python-objects)

jasondet:
@iguinn we are ready for you to revisit this

PR #252, opened 2022-04-19
Fixed numba version check

iguinn:
Python's setup tools do not for subversions, so it does not recognize numba version `0.53.1` as breaking `!=0.53`. Fixed by explicitly adding `0.53.1` and other sub-versions of `0.53` and `0.54`.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -145,7 +145,7 @@ setup(
     install_requires=[
         'numpy',
         'scimath',
-        'numba !=0.53, !=0.54',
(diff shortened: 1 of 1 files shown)


gipert:
This unfortunately still does not catch versions strings like `0.53.0rc1`, however I just discovered that apparently one can write something like `!=0.53.*`. I'm updating this in the refactor of the package configs.

iguinn:
Ah, nice find!

PR #255, opened 2022-04-25
Address some testing and packaging issues

gipert:
Lots of new features: here's a short summary of the changes.

- [x] Enable GitHub [Dependabot](https://docs.github.com/en/code-security/dependabot/working-with-dependabot/keeping-your-actions-up-to-date-with-dependabot) to keep GitHub actions automatically up to date. There will be a bot opening a pull request once in a while (I expect it to be very quiet) to update the GitHub action workflow file.
- [x] Add https://codecov.io configuration file (this will enable the corresponding bot) and upload coverage automatically with GitHub actions. For every pull request the Codecov bot will post a summary of the test coverage status and tell us whether the proposed changes are sufficiently tested. Let's see how noisy this is: we can temporary disable the bot while in this heavy-development phase
- [x] Enable the https://pre-commit.ci bot to automatically perform style checks through [pre-commit](https://pre-commit.com): there will be now an additional check run for every commit ensuring that coding conventions are respected. The pre-commit.ci bot will automatically fix (wherever possible) style-related issues in every PR, by adding a commit. The list of checks to be performed are specified through the [.pre-commit-config.yaml](https://github.com/legend-exp/pygama/blob/158f04b3dd0b55f19f5b346a9cd8db08282d8bed/.pre-commit-config.yaml) configuration file. At the moment I just enabled some basic (but extremely useful!) checks: trailing whitespace or newlines, dead symlinks, large files... Later we can also enable some python linters (like black or flake8) or static type checkers (mypy): this will be also extremely useful but requires a massive re-format of the existing code – we can do it once the refactoring is over
- [x] Update developer docs

See also #241 and #242.

PR #257, opened 2022-04-27
Switch to declarative package management

gipert:
I've been mainly following recommendations from here: https://scikit-hep.org/developer/packaging

* packaging instructions are now contained in `pyproject.toml`, `setup.cfg` (and few lines in `setup.py` to be backward-compatible)
* package source has been moved below `src/` as recommended: simplifies packaging by a lot
* the pygama version is now managed by setuptools, which computes a `git-describe`-like version string automatically. I didn't know about this before and I consider it robust enough (requires just a couple of extra configs in `pyproject.toml` and no extra custom logic). We will not need to manually advance the version string anymore.
* export only pygama version string as LH5 object

Still to do:
- [x] update after #255 is merged
- [x] update contributing guide
- [x] update GHA workflow
- [x] add missing optional dependencies (such as pyfcutils)

@jasondet: are you OK with moving the source below `src/`? Do you have any particular development requirement/need?

**To be merged after #255**


gipert:
I'm done with this PR – let me know what you think @jasondet @iguinn and feel free to merge, if no objections. https://github.com/legend-exp/pyfcutils/pull/2 should be merged too to make fcutils install seamlessly.

jasondet:
oh wow. this looks like a good idea to me. I've requested that @iguinn and @wisecg review it.

wisecg:
Wow!  I'm impressed at the work here.  I support any change that makes pygama more aligned with best practices in ScikitHEP.   Eventually it would be great to publish this package with them.

The only thing I note here is that we would move the "pygama" dir one level down, to "src/pygama".   As long as Luigi feels that's necessary, I'll support his decision.

wisecg (approved):

gipert:
> Wow! I'm impressed at the work here. I support any change that makes pygama more aligned with best practices in ScikitHEP. Eventually it would be great to publish this package with them.

I guess it would be nice to contact them at some point, yes: https://scikit-hep.org/affiliated. But ultimately it will depend on pygama's ability to be not-so-much LEGEND-specific.

> The only thing I note here is that we would move the "pygama" dir one level down, to "src/pygama". As long as Luigi feels that's necessary, I'll support his decision.

Yes, it's generally highly recommended as most tools assume this structure by default (and therefore no extra configuration is needed): https://scikit-hep.org/developer/pep621#package-structure

jasondet:
ok. I'm going to merge this...

PR #258, opened 2022-05-02
LGDO str() 

gracesong312:
The biggest change is to let WaveformTable create a VectorOfVectors if it receives an array of lists as 'values'

The strings themselves should be pretty straightforward, the only one I'm worried about is Table - I return the string of get_dataframe(), but the dataframe is only useful if the data is an Array or VectorOfVectors (i.e. a list or an np array don't have 'nda' attributes)


pygama/lgdo/table.py
```diff
@@ -11,7 +13,7 @@ class Table(Struct):
     table I/O using functions like push_row(), is_full(), and clear()
 
     Note: if you write to a table and don't fill it up to its total size, be
-    sure to resize it before passing to data processing functions, as they will
+    sure to resize it before passing to data processing fucntions, as they will
```
jasondet, line 16: typo here

pygama/lgdo/table.py
```diff
@@ -171,7 +173,17 @@ def get_dataframe(self, cols=None, copy=False):
         df = pd.DataFrame(copy=copy)
         if cols is None: cols = self.keys()
         for col in cols:
-            if not hasattr(self[col],'nda'):
+            if isinstance(self[col], VectorOfVectors):
+                df[col] = self[col].tolist()
```
jasondet, line 177: @wisecg will this work? Will pandas make a dataframe from a list of lists of possibly unequal length?

pygama/lgdo/vectorofvectors.py
```diff
@@ -47,7 +47,9 @@ def __init__(self, flattened_data=None, cumulative_length=None, shape_guess=None
             length = np.prod(shape_guess)
             self.flattened_data = Array(shape=(length,), dtype=dtype)
         else: self.flattened_data = flattened_data
-        self.dtype = self.flattened_data.dtype
+        if dtype is None:
+            self.dtype = self.flattened_data.dtype
+        else: self.dtype = dtype
```
jasondet, line 52: Oh this is an interesting block. There are 4 options:
- flattened_data is None and dtype is not None: this is supposed to not be allowed; presumably Array's __init__ function should emit a warning in this case, but it looks like it doesn't. I think this was already a problem in the original code that should probably be fixed.
- flattened_data is None and dtype is not None: make an empty array with the requested dtype (ok in both versions)
- flattened data is not None and dtype is None: use the input array and its keep its dtype. This is correct in both versions.
- flattened data is not None and dtype is not None: use the input array but change its dtype if necessary. This appears to not be implemented in this way in both the new and the old code.

pygama/lgdo/vectorofvectors.py
```diff
@@ -94,3 +96,20 @@ def set_vector(self, i_vec, nda):
             self.flattened_data.nda.resize(2*len(self.flattened_data.nda))
         self.flattened_data.nda[start:end] = nda
         self.cumulative_length.nda[i_vec] = end
+
+    def tolist(self):
+        l = []
+        l.append(list(self.flattened_data[:self.cumulative_length[0]]))
+        l += [list(self.flattened_data[self.cumulative_length[i-1]:self.cumulative_length[i]]) for i in range(1, len(self.cumulative_length))]
```
jasondet, line 103: This uses an explicit loop, which could be slow. I wonder if implementing __iter__() might be better. I'm also not remembering why we didn't use awkward arrays for VofV

wisecg:
Jason and I discussed this PR this morning.   My suggestion is that we don't change the way the LGDO's are implemented on disk -- the only two dependencies needed at this point are h5py and numpy.    What we would like are features which read in LGDO objects using only h5py and numpy -- then if a user would like, we can return a _representation_ of that data in a more standard format such as pandas or awkward-array.    

The `get_dataframe` function is going to work well when the data is a flat table, such as ArrayOfEqualSizedArrays.    Pandas isn't good when each row has a variable length, so for something like VectorOfVectors, we should have a function called `get_awkward_array` which can convert LH5 on disk to an in-memory awkward-array.

gipert:
Hi @gracesong312, thanks for fixing the problems reported by pre-commit hooks. It's unfortunately still failing because of #260 – you should poke @jasondet for that ;)

jasondet:
@gracesong312 this is very good, especially the important updates to vov. I think there's one last thing -- even though vov can be converted nicely into a list now, I think we want table.get_dataframe() to still not allow users to specify a vov field. We'll want them to manually change it to an aoesa or similar. If you fix that one thing I think this will be ready to merge!

If you get a chance, please also add a unit test for list(vov)

src/pygama/lgdo/table.py
```diff
@@ -153,7 +155,7 @@ def join(self, other_table, cols=None, do_warn=True):
     def get_dataframe(self, cols=None, copy=False):
         """Get a pandas dataframe from the data in the table.
 
-        Note: the requested data must be array-like, with the nda attribute.
+        Note: the requested data must be array-like, with the nda attribute, or a VectorOfVectors.
```
jasondet, line 158: I think the "or a VoV" comment here needs to be removed now?

src/pygama/lgdo/table.py
```diff
@@ -173,5 +175,6 @@ def get_dataframe(self, cols=None, copy=False):
         for col in cols:
             if not hasattr(self[col],'nda'):
                 print(f'column {col} does not have an nda, skipping...')
-            else: df[col] = self[col].nda
+            else: df[col] = list(self[col].nda)
```
jasondet, line 178: I think the conversion to a list is unnecessary, right? It should be removed because it takes CPU + memory.

jasondet:
Nice! Thanks Grace!

jasondet:
oh wait. now there are checks to fix 😅

jasondet:
thx!

PR #260, opened 2022-05-18
orca header handling

jasondet:
Refactor


gipert:
Hey @jasondet, does this complete the refactor of the orca module? I see there are some merge conflicts committed https://github.com/legend-exp/pygama/blob/refactor/src/pygama/raw/orca/orca_base.py#L1-L18

PR #261, opened 2022-05-23
Updated ORCA struck 3316 decoder to handle channel groups. 

rbbouabid:
Required dealing w/ bug in memory in the set_object_info function.

The solution that we came up with was to import the python package copy and make a deep copy of the decoded_values_template, because as the template was being updated in the channel loop every channel's decoded_value_template was changing, thus defeating the purpose of the whole exercise.

If someone has a better idea for how to accomplish w/out introducing a new dependency on the package "copy" we're open to another solution. Otherwise this seems to work inside of the LEGEND-base container.


jasondet:
This is the right solution. will implement it for other decoders.

PR #264, opened 2022-05-24
`raw.fc` unit testing

gipert:
Including a couple of bug fixes. I'm not always 100% sure I'm testing the right things, so let me know. Let's see how test coverage improves.


gipert:
@jasondet @gracesong312: looks like the latest changes in `VectorOfVectors` are breaking the FlashCam decoders, can you please check and push a fix?

https://github.com/legend-exp/pygama/runs/6625033967?check_suite_focus=true#step:5:67

jasondet:
@gracesong312 can you take a look?

gracesong312:
These should be fixed by #265

gracesong312:
So I think this was an existing problem, but what's happening is when an argument in WaveformTable is not specified, it defaults to None but the fields 'dt', 't0', and 'values' are still set. In this case, we have an entry in the table {'values': None} and when resize gets called by the constructor and starts looping through items, it's trying to call len(None) and failing. 

What do we want the behavior to be when size is specified but another field is not?

gipert:
@jasondet shouldn't it be enough if we add a `is not None` check in `table.resize`?

gipert:
Why is `values` still `None` after this?

https://github.com/legend-exp/pygama/blob/refactor/src/pygama/lgdo/waveform_table.py#L91-L116

gracesong312:
Because None is not an instance of `np.ndarray` so it skips that whole block

gipert:
I guess that if `values` is `None`, we should initialize it with the available info (see the docstring). Is not enough info is provided, we should throw an exception (?)

gipert:
The problem is here @gracesong312:

https://github.com/legend-exp/pygama/pull/258/files#diff-e7585c39741e9081221c659a37917a2c922b628716cc7e9f0541139b96c71505

Before the additions, a `None` `values` was being correctly initialized before being given to the `Table` constructor. Now the code is inside `if isinstance(values, np.ndarray):`, so `values == None` is not catched anymore.

gipert:
The problem is here @gracesong312:

https://github.com/legend-exp/pygama/pull/258/files#diff-e7585c39741e9081221c659a37917a2c922b628716cc7e9f0541139b96c71505

Before the additions, a `None` `values` was being correctly initialized before being given to the `Table` constructor. Now the code is inside `if isinstance(values, np.ndarray):`, so `values == None` is not catched anymore.

gracesong312:
I'm having issues getting `lgnd_test_data` to work to run your tests, but the `decoder.make_lgdo(size=1)` line runs. The only change I made is deleting one indent on the block that you mentioned from lines 97 to 115 in `waveform_table.py` 

Would it be easier for you to just make that change instead of me opening another pull request without being able to fully test it?

gipert:
No problem, I can push the fix here.

gipert:
Pygama needs `python>=3.9`: https://github.com/legend-exp/pygama/runs/6686665617?check_suite_focus=true#step:5:275

We should quickly discuss about how to proceed today at the call. In my opinion, we should make a survey of the Python3 versions currently pre-installed on most common platforms (OSX, Linux distros, containers...) and unsupport older Python versions, if possible.

cc @jasondet @iguinn @wisecg

jasondet (approved):
These tests look great.

jasondet:
@gipert just figure out what's going on with the untar failure and then we can merge.

PR #269, opened 2022-06-02
Fix timestamp calculation.

ssailer:
Fixes a typo in the timestamp calculation.


jasondet:
oh wow good catch, thank you!

PR #276, opened 2022-06-08
fix lh5_store tests

jasondet:


gipert:
Looks great! Thanks for fixing.

PR #278, opened 2022-06-08
Various documentation improvements

gipert:
@jasondet I have fixed a few things in `raw.orca` that were causing troubles to Sphinx (I am aware those would have been anyways fixed with the refactor). I have configured now the Sphinx build (on GitHub actions too) to fail if the documentation cannot be rendered properly, so we can make sure that broken docs will not be deployed to the web.


jasondet:
ok by me. thanks Luigi

PR #281, opened 2022-06-08
First iteration of math refactor

SamuelBorden:
Major changes include: 

- splitting fitting routines from `peak_fitting.py` into  `linear_fitting.py, binned_fitting.py, unbinned_fitting.py`
- Mathematical functions from `peak_fitting.py` were split into `functions.py` and then a subpackage `distributions.py`. The latter contains PDFs and CDFs grouped in their own files, this could eventually be turned into classes. 
- Removed `peak_fitting.fit_hist` which is replaced by `binned_fitting.fit_binned` 
- Removed `utils.sh`
- Removed `peak_fitting.cal_slope`, will be replaced in the pargen refactor with the correct function.


gipert:
Thanks a lot for the work @SamuelBorden! I have a few comments.

- Usually, in Python packages, prepending subpackage names with `_` is used to label "private" code. Private means that the code is not meant to be used directly by the pygama user (i.e. is not part of the public API). In this case, I would rather call the subpackage `distributions` and ` mv distributions.py distributions/__init__.py`, so one can still do `import distributions.double_gauss_pdf`. We have the underscore in `dsp._processors` because processors are actually not meant to be used directly (maybe @iguinn can comment here).
- Why not calling the entire subpackage `stats`, instead of `math`?
- I see we define ourselves many utility functions like the Gauss pdf. Why don't we use some third party definitions here? Is it just because we need them to be Numbafy?
- Can we use a different name for what is called `radford_*`? For example, what about `hpge_peak_pdf` instead of `radford_pdf`?

src/pygama/math/_distributions/double_gauss.py
```diff
@@ -0,0 +1,94 @@
...
+"""
+import math
+
+import numpy as np
+
+from pygama.math._distributions.gauss import gauss_norm
+from pygama.math._distributions.step import step_pdf
+
+
+def Am_double(x,  n_sig1, mu1, sigma1,  n_sig2, mu2,sigma2, n_sig3, mu3,sigma3, n_bkg1, hstep1, n_bkg2, hstep2,
```
gipert, line 12: I find these names for the Am utilities a bit cryptic, maybe `am241_double_peak_pdf`?

jasondet: rename it. something like double_gauss_step

src/pygama/math/_distributions/double_gauss.py
```diff
@@ -0,0 +1,94 @@
...
+    sig1 = n_sig1*gauss_norm(x,mu1,sigma1)
+    sig2 = n_sig2* gauss_norm(x,mu2,sigma2)
+    sig3 = n_sig3* gauss_norm(x,mu3,sigma3)
+    if components ==False:
+        return sig1+sig2+sig3+bkg1+bkg2
+    else:
+        return sig1,sig2,sig3,bkg1,bkg2
+
+
+def extended_Am_double(x,  n_sig1, mu1, sigma1,  n_sig2, mu2,sigma2, n_sig3, mu3,sigma3,
```
jasondet, line 35: rename (triple_exgauss_step?)
--> later: make a class that generically combines multiple pdfs into a single function

src/pygama/math/_distributions/double_gauss.py
```diff
@@ -0,0 +1,94 @@
...
+                                                               n_bkg1, hstep1, n_bkg2, hstep2,
+                                                                 lower_range, upper_range)
+    else:
+        sig1,sig2,sig3,bkg1,bkg2 = Am_double(n_sig1, mu1, sigma1,  n_sig2, mu2,sigma2, n_sig3, mu3,sigma3,
+                                             n_bkg1, hstep1, n_bkg2, hstep2,
+                                             lower_range , upper_range,components=components)
+        return n_sig1+n_sig2+n_sig3 + n_bkg1+n_bkg2, sig1,sig2,sig3,bkg1,bkg2
+
+
+def double_gauss_pdf(x,  n_sig1,  mu1, sigma1, n_sig2, mu2,sigma2,n_bkg,hstep,
```
jasondet, line 58: remove (same as double_gauss_step with 2nd step fixed to 0)

src/pygama/math/_distributions/double_gauss.py
```diff
@@ -0,0 +1,94 @@
...
+        return 0, np.zeros_like(x)
+    sig1 = n_sig1*gauss_norm(x,mu1,sigma1)
+    sig2 = n_sig2* gauss_norm(x,mu2,sigma2)
+    if components == False:
+        return sig1 + sig2 + bkg
+    else:
+        return sig1, sig2, bkg
+
+
+def extended_double_gauss_pdf(x,  n_sig1,  mu1, sigma1, n_sig2, mu2,sigma2,n_bkg,hstep,
```
jasondet, line 78: remove

src/pygama/math/_distributions/gauss.py
```diff
@@ -0,0 +1,40 @@
+"""
+Gaussian distributions for pygama
+"""
```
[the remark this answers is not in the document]
jasondet: later: put scipy function / distribution base classes under these?

src/pygama/math/_distributions/gauss_on_background.py
```diff
@@ -0,0 +1,30 @@
+"""
+Gaussian distributions on linear and uniform backgrounds for pygama
+"""
+import numpy as np
+
+from pygama.math._distributions.gauss import gauss_norm
+
+
+def gauss_uniform(x, n_sig, mu, sigma, n_bkg, components = False):
```
jasondet, line 9: candidate for removal: make gauss_poly

src/pygama/math/_distributions/gauss_step.py
```diff
@@ -0,0 +1,63 @@
...
+        bkg= np.zeros_like(x, dtype=np.float64)
+    if components ==False:
+        cdf = (1/(n_sig+n_bkg))*(n_sig*gauss_cdf(x, mu, sigma) +\
+          n_bkg*bkg)
+        return cdf
+    else:
+        return (1/(n_sig+n_bkg))*n_sig*gauss_cdf(x, mu, sigma), (1/(n_sig+n_bkg))*(n_bkg*bkg)
+
+
+def extended_gauss_step_pdf(x,  n_sig, mu, sigma, n_bkg, hstep, lower_range=np.inf , upper_range=np.inf, components=False):
```
jasondet, line 54: just call the non-pdf/cdf version "gauss_step"

src/pygama/math/_distributions/gauss_with_tail.py
```diff
@@ -0,0 +1,121 @@
...
+        integral = gauss_tail_integral(np.array([np.nanmin(x), np.nanmax(x)]), mu, sigma, tau)
+    else:
+        integral = gauss_tail_integral(np.array([lower_range, upper_range]), mu, sigma, tau)
+    norm = integral[1]-integral[0]
+    cdf =  (1/norm) * cdf
+    c = 1-cdf[-1]
+    return cdf+c
+
+
+def gauss_with_tail_pdf(x, mu, sigma,  htail,tau, components=False):
```
jasondet, line 46: remove these?

src/pygama/math/_distributions/gauss_with_tail.py
```diff
@@ -0,0 +1,121 @@
...
+
+from pygama.math._distributions.gauss import gauss_cdf, gauss_norm
+from pygama.math.functions import gauss_tail_exact, nb_erf
+
+limit = np.log(sys.float_info.max)/10
+kwd = {"parallel": False, "fastmath": True}
+
+
+@nb.njit(**kwd)
+def gauss_tail_pdf(x, mu, sigma, tau):
```
jasondet, line 17: rename exgauss

SamuelBorden:
I've implemented most comments from above. However, I left some major functional improvements for later. A future PR could create the base function class, and then implement the methods needed for distribution selection. Likewise, a future PR can remove `fit_binned` and `fit_unbinned`. Other future work could include unit tests for all of the distributions, as well as adding type annotations as @gipert suggested. I ended up not removing `polynomial.py` because code in pargen calls it frequently.

src/pygama/math/histogram.py
```diff
@@ -125,23 +128,23 @@ def better_int_binning(x_lo=0, x_hi=None, dx=None, n_bins=None):
     # process inputs
     n_Nones = int(x_hi is None) + int(dx is None) + int(n_bins is None)
     if n_Nones > 1:
-        print('better_int_binning: must provide two of x_hi, dx or n_bins')
-        return
+        log.warning('better_int_binning: must provide two of x_hi, dx or n_bins')
+        raise Exception
```
gipert, line 132: Some notes about errors/warnings:
* do not add the current function name, it is automatically added by logging / stacktrace
* Is it a warning? Then just call `log.warning()` and don't raise an exception
* Is it an error (e.g. execution cannot continue)? Then just raise an exception ([the one that makes most sense!](https://docs.python.org/3/library/exceptions.html#built-in-exceptions)) and pass the nessage to it.

You would change the above with:
```suggestion
        raise ValueError('must provide two of x_hi, dx or n_bins')
```
Maybe `ValueError` is not exactly the right one. In that case you would throw something more generic like `RuntimeError`.

Please update the rest of the PR accordingly.

gipert:
Uhm why are `build_raw` tests failing... odd

jasondet:
ok going to finally merge this for CJ. Sam will continue refactoring math.

PR #282, opened 2022-06-10
Add more high-level `raw` tests

gipert:
I replaced some `print('error'); return` with exception raising. We should try to do this more consistently.


jasondet:
Very nice.

I totally agree on switching to exception raising -- I'll get in the habit.

PR #283, opened 2022-06-22
Use `logging` module in `pygama.raw`, add CLI

gipert:
* Verbosity in `raw` is now managed through the logging module, allowing for a much advanced user configuration. Will do the same for the rest of pygama in future PRs
* Add command line interface, accessible through the `pygama` shell command. Available for `build_raw` only at the moment. Will add `build_dsp` etc in future PRs
* `pygama --verbose ...` increases the verbosity level. In user scripts, verbosity must be configured by the user via the `logging` module, for example:
```py
import logging
logging.basicConfig(level=logging.DEBUG, format="%(name)s [%(levelname)s] %(message)s")

import pygama
...
```

### DIFF
--- a/setup.cfg
+++ b/setup.cfg
@@ -8,6 +8,7 @@ author = The LEGEND collaboration
 maintainer = The LEGEND collaboration
 license = Apache-2.0
 license_file = LICENSE
+license_files = LICENSE
 classifiers =
     Development Status :: 4 - Beta
     Intended Audience :: Developers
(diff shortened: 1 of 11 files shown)


jasondet:
That logger looks awesome, it really cleans up the code! Thank you!

On the CLI -- it uses command names like "daq2raw", can you change them to all be like "build_raw" like we use in the refactor?

jasondet:
nice. thx!

PR #285, opened 2022-06-23
Add some first high-level ORCA decoding tests

gipert:
Test decoding May ORCA data from legend-testdata. Tests should now pass after #284.

### DIFF
--- a/src/pygama/raw/build_raw.py
+++ b/src/pygama/raw/build_raw.py
@@ -81,7 +81,7 @@ def build_raw(in_stream, in_stream_type=None, out_spec=None, buffer_size=8192,
             ext = in_stream.split('/')[-1][i_ext+1:]
             if ext == 'fcio':
                 in_stream_type = 'FlashCam'
-            elif ext == 'gz' and OrcaStreamer.is_orca_stream(in_stream):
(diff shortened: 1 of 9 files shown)


src/pygama/raw/build_raw.py
```diff
@@ -81,7 +81,7 @@ def build_raw(in_stream, in_stream_type=None, out_spec=None, buffer_size=8192,
             ext = in_stream.split('/')[-1][i_ext+1:]
             if ext == 'fcio':
                 in_stream_type = 'FlashCam'
-            elif ext == 'gz' and OrcaStreamer.is_orca_stream(in_stream):
+            elif ext == 'orca' and OrcaStreamer.is_orca_stream(in_stream):
```
jasondet, line 84: there are definitely orca files on other systems with names like Run0030.gz, so please don't remove .gz

gipert: Why not just calling `is_orca_stream` then?

jasondet: is_orca_stream reads 12 binary bytes and checks if they look like they come from Orca. I guess I was worried about a non-orca file randomly passing the test. but for truly random data the chances of that are like 1/2^100. So Yeah -- I think we can just call is_orca_stream without checking the extension.

PR #287, opened 2022-06-29
Updated moving window and added upsampler

ggmarshall:
Changes to initial definitions of the moving windows due to issues found. Also added upsampler which is used for the A/E determination


gipert:
Yep, pre-commit is not configured in `main`, you can merge. We can then open a PR `main` -> `refactor` and see what happens.

PR #288, opened 2022-06-29
Friendly `build_raw` command line interface

gipert:
I implemented a special keyword `orig_basename` that gets substituted in the config dictionary by the name of the original DAQ file (without extension). This makes it easier to quickly convert DAQ files on the command line. Example:
```console
$ cat orca-out-spec-cli.json
{
    "ORFlashCamADCWaveformDecoder": {
        "g{key}": {
            "key_list": [[0, 3]],
            "out_stream": "/tmp/{orig_basename}.lh5"
        }
    }
}
$ pygama build-raw -w -o orca-out-spec.json legend-testdata/data/orca/fc/L200-comm-20220519-phy-geds.orca
pygama.raw.build_raw [INFO] time elapsed: 0.04 sec
pygama.raw.build_raw [INFO] output file: /tmp/L200-comm-20220519-phy-geds.lh5 (682.289 KB)
pygama.raw.build_raw [INFO] total converted: 988.125 KB
pygama.raw.build_raw [INFO] conversion speed: 26.420 MBps
```
What do you think @jasondet? There was no way to re-use the original filename in `out_stream` otherwise.

### DIFF
--- a/src/pygama/cli.py
+++ b/src/pygama/cli.py
@@ -1,7 +1,7 @@
 import argparse
 import logging
-import sys
 import os
+import sys
 
 import numpy as np
 
--- a/src/pygama/cli.py
+++ b/src/pygama/cli.py
@@ -1,6 +1,7 @@
 import argparse
 import logging
 import sys
+import os
 
 import numpy as np
 
@@ -15,9 +16,9 @@ def pygama_cli():
         description="pygama's command-line interface")
 
     # global options
-    parser.add_argument('--version', '-v', action='store_true',
+    parser.add_argument('--version', action='store_true',
                         help="""Print pygama version and exit""")
-    parser.add_argument('--verbose', action='store_true',
+    parser.add_argument('--verbose', '-v', action='store_true',
                         help="""Increase the program verbosity""")
 
     subparsers = parser.add_subparsers()
@@ -29,11 +30,11 @@ def pygama_cli():
                             help="""Input stream. Can be a single file, a list
                             of files or any other input type supported by the
                             selected streamer""")
-    parser_d2r.add_argument('--stream-type', '-t', nargs=1,
+    parser_d2r.add_argument('--stream-type', '-t',
                             help="""Input stream type name. Use this if the
                             stream type cannot be automatically deduced by
                             pygama""")
-    parser_d2r.add_argument('--out-spec', '-o', nargs=1,
+    parser_d2r.add_argument('--out-spec', '-o',
                             help="""Specification for the output stream. HDF5
                             or JSON file name""")
     parser_d2r.add_argument('--buffer_size', '-b', type=int, default=8192,
@@ -62,6 +63,10 @@ def pygama_cli():
     if args.verbose:
         logging.basicConfig(level=logging.DEBUG,
                             format="%(name)s [%(levelname)s] %(message)s")
+    else:
+        logging.basicConfig(level=logging.INFO,
+                            format="%(name)s [%(levelname)s] %(message)s")
+
     if args.version:
         print(pygama.__version__)
         sys.exit()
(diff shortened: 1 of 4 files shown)


src/pygama/cli.py
```diff
@@ -71,9 +76,11 @@ def pygama_cli():
 
 def build_raw_cli(args):
     for stream in args.in_stream:
+        basename = os.path.splitext(os.path.basename(stream))[0]
         build_raw(stream, in_stream_type=args.stream_type,
                   out_spec=args.out_spec, buffer_size=args.buffer_size,
-                  n_max=args.max_rows, overwrite=args.overwrite)
+                  n_max=args.max_rows, overwrite=args.overwrite,
+                  orig_basename=basename)
```
gipert, line 83: This is how I do it.

jasondet:
the json for the raw buffer configuration already has a generic method of doing keyword substitutions, see the paragraph right above the sample json here:
https://legend-exp.github.io/pygama/refactor/generated/pygama.raw.html#primary-classes

Here's the python script I'm using on NERSC for generating data that makes use of this feature:
```
import sys, os, json
from pygama.raw.build_raw import build_raw
from pygama.raw.orca.orca_streamer import OrcaStreamer

if len(sys.argv) < 2:
    print(f"Usage: python {sys.argv[0]} [files]")
    print(f"e.g.: python {sys.argv[0]} /global/project/projectdirs/legend/data/lngs/l200_commissioning/orca_data/Data/p0/r2/L200_p0_r2_20220606T151738Z_f0.orca")
    sys.exit()

out_spec_json = '''
{
  "ORFlashCamADCWaveformDecoder" : {
    "ch{key:03d}/raw" : {
      "key_list" : ["*"],
      "out_stream" : "{filekey}.lh5"
    }
  }
}
'''

#for arg in sys.argv[1:]:
for arg in sys.argv[9:]:
    daq_file_path = arg

    if not os.path.exists(daq_file_path):
        print(f"File {sys.argv[1]} doesn't exist")
        sys.exit()

    filekey = os.path.basename(daq_file_path)
    dot_pos = filekey.rfind('.')
    if dot_pos >= 0: filekey = filekey[:dot_pos]
    if filekey.startswith('L200'): filekey = 'L60'+filekey[4:]

    build_raw(daq_file_path, in_stream_type='ORCA', out_spec=json.loads(out_spec_json), filekey=filekey)
```

We often can't re-use the DAQ name as-is, so something more generic is typically required than just using the input basename.

jasondet:
I think that functionality is already available by calling `build_raw()` without giving it an `out_spec` argument, with the exception that I think in that case it writes the output to the same directory as the input files, but we can change that. So someone would just do

`$ pygama build-raw data/*.orca`

which I think is even nicer ;)

jasondet:
I think I'm missing your argument. Can you explain in more detail? My design was:
- with no arguments, build_raw() generates an output file with the same name as the input file, with an HDF5 group for each decoder, containing a table of data for that decoder
- if users want a different output filename or HDF5 group layout, they do so by providing a json config to the out_spec argument.
- if the json config has case-specific strings, the user specifies each with a keyword of their choice, then sends in the value for each specific case in the kwargs of build_raw()

This looked to me like a codification of a specific kwarg so I thought maybe it's not necessary. But if it's providing new functionality then I'm happy to include it.

gipert:
It is indeed a codification of a specific kwarg (with the CLI only!!!), but how do you set kwargs from the command line otherwise? We could implement something like `pygama build-raw --kwargs "var1=val1;var2=val2"`...

jasondet:
oh I'm an idiot -- i totally missed that this was in the cli only! I thought you were baking it into build_raw() itself.

I agree this is really nice for the CLI! I'm going to merge it.

PR #289, opened 2022-06-29
Add type annotations to important `raw` methods and improve docstrings

gipert:


src/pygama/raw/build_raw.py
```diff
@@ -17,8 +19,9 @@
 log = logging.getLogger(__name__)
 
 
-def build_raw(in_stream, in_stream_type=None, out_spec=None, buffer_size=8192,
-              n_max=np.inf, overwrite=True, **kwargs):
+def build_raw(in_stream: int, in_stream_type: str = None,
+              out_spec: str | dict | RawBufferLibrary = None, buffer_size: int = 8192,
+              n_max: int = np.inf, overwrite: bool = True, **kwargs) -> None:
```
jasondet, line 24: np.inf is a float -- is it okay to say n_max must be an int and give it a float default value? Or is there an int max that we can change it to?

src/pygama/raw/data_decoder.py
```diff
@@ -46,50 +48,49 @@ def __init__(self, garbage_length=256, packet_size_guess=1024):
...
+            ``decoded_values``. Otherwise, just returns
+            :attr:`self.decoded_values`, which should be defined in the
+            constructor.
         """
         if key is None:
             return self.decoded_values if hasattr(self, 'decoded_values') else None
 
         raise NotImplementedError("you need to implement key-specific get_decoded_values for", type(self).__name__)
 
+    def make_lgdo(self, key: Any = None, size: int = None):
```
jasondet, line 72: same question here on setting the int `size` to `None`

src/pygama/raw/data_streamer.py
```diff
@@ -31,19 +34,21 @@ def __init__(self):
         self.packet_id = 0
 
 
-    def open_stream(self, stream_name, rb_lib=None, buffer_size=8192,
-                    chunk_mode='any_full', out_stream=''):
-        """ Open and initialize a data stream
+    def open_stream(self, stream_name: str, rb_lib: RawBufferLibrary = None,
```
jasondet, line 37: same question here

src/pygama/raw/data_streamer.py
```diff
@@ -152,29 +154,29 @@ def read_packet(self):
         """
         return True
 
-
-    def read_chunk(self, chunk_mode_override=None, rp_max=1000000, clear_full_buffers=True):
-        """
-        Reads a chunk of data into raw buffers
+    def read_chunk(self, chunk_mode_override: str = None, rp_max: int = 1000000,
```
jasondet, line 157: and here

jasondet: (I'll stop marking these from here)

jasondet:
Oh wow thank you @gipert I was hoping to start adding typing "some day" and you already got it started! See my few comments -- if they are non-issues, feel free to merge.

src/pygama/raw/data_decoder.py
gipert: Since `None` is already specified as default value, writing `size: int | None = None` seems a bit redundant to me. My philosophy is that, since type annotations are not checked by the interpreter, we should try to privilege readability for the user...

We might revise all this once (and if...) we'll have static type checkers like MyPy running as part of our CI.

gipert: Since we're on this function declaration: should we be more restrictive than `Any` for the key type? What would you put there?

jasondet: Sounds good on not working about int vs None.

As long as its just a suggestion to the user, I would propose `int | str`. In principle a user could use anything, but we want to be able to key dictionaries and name HDF5 datasets using `key`, so as a suggestion to the user and int or a str is heavily preferred over other options.

jasondet:
Ok -- decide if you want to change `Any` to `int | str` above, then please merge.

PR #310, opened 2022-07-06
`dsp` tests

gipert:
- added some first tests
- fix/add type hints
- fix docstrings
- make overwrite mode = false in build_raw and build_dsp


jasondet:
@gipert when you get a chance, can you solve the conflict? Also, shouldn't this PR be on main not refactor?

jasondet:
@iguinn can you review this when you get a chance?

tests/dsp/test_build_dsp.py
```diff
@@ -0,0 +1,77 @@
...
+    }
+
+    build_raw(in_stream=lgnd_test_data.get_path('fcio/L200-comm-20211130-phy-spms.fcio'),
+              out_spec=out_spec, overwrite=True)
+
+    return out_file
+
+
+def test_build_dsp_basics(lgnd_test_data):
+    build_dsp(lgnd_test_data.get_path('lh5/LDQTA_r117_20200110T105115Z_cal_geds_raw.lh5'),
```
iguinn, line 34: I think I'm missing something here...What kind of object is `lgnd_test_data` and how is it getting passed here? I assume this is pointing to the test data repository?

tests/dsp/configs/icpc-dsp-config.json
```diff
@@ -0,0 +1,327 @@
+{
```
iguinn, line 1: Are we sure we want to use a full-fledged DSP file like this, rather than a more minimal example? It seems like the job of testing the individual processors should go to unit_tests, and for testing build_dsp and processing_chain, we may want a more minimal file

iguinn (commented):
As someone fairly inexperienced in writing tests for python, I want to make sure that I understand the goals for each test:
test_processing_chain.py: this seems to be essentially have integration tests to make sure that it can run the kinds of processors we want it to run (like running a scipy processor) and unit tests to make sure that the parsing of different cases works
test_build_dsp.py: this seems to be a functional test that's designed to make sure that we can process a real file
test_waveform_browser.py: this seems to be a functional test that draws waveforms

For test_build_dsp, as I commented above, do we want this to use a full-fledged LEGEND DSP config file, or do we want to pare it down to more of a minimal working file? My (possibly flawed) understanding is that we would prefer to use unit testing for the individual processors and design this test to be simpler?

For test_waveform_browser.py, this involves actually drawing waveforms. How does this work for automatic testing? Is it more or less just going to look for if any errors are thrown, or is this more meant for someone to run on their own to visually confirm that waveforms are drawn correctly?

jasondet:
@iguinn can you check these conflicts when you get a chance?

tests/dsp/test_build_dsp.py
```diff
@@ -0,0 +1,77 @@
...
+from pathlib import Path
+
+import pytest
+
+from pygama import lgdo
+from pygama.dsp import build_dsp
+from pygama.lgdo.lh5_store import LH5Store, ls
+from pygama.raw import build_raw
+
+config_dir = Path(__file__).parent/'configs'
```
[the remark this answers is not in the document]
gipert: I don't think it's useful t to install test data/configs. By design, the test suite is meant to be run by developers on the pygama source (not the installed code!).

For tutorials/docs the story is different. I agree that we should include them as installed data files.

tests/dsp/test_build_dsp.py
gipert: It's defined [here](https://github.com/legend-exp/pygama/blob/main/tests/conftest.py#L5-L9) as a [Pytest fixture](https://docs.pytest.org/en/7.1.x/how-to/fixtures.html). Fixtures are useful to share objects between tests.

tests/dsp/configs/icpc-dsp-config.json
gipert: I agree that we want unit tests for the processors folder. My plan (not part of this PR) is to have a test for each single processor. Still, testing something similar to what we have in the L200 data production seems useful to me...

gipert:
>As someone fairly inexperienced in writing tests for python, I want to make sure that I understand the goals for each test:
test_processing_chain.py: this seems to be essentially have integration tests to make sure that it can run the kinds of processors we want it to run (like running a scipy processor) and unit tests to make sure that the parsing of different cases works
test_build_dsp.py: this seems to be a functional test that's designed to make sure that we can process a real file
test_waveform_browser.py: this seems to be a functional test that draws waveforms

Correct.

For test_build_dsp, as I commented above, do we want this to use a full-fledged LEGEND DSP config file, or do we want to pare it down to more of a minimal working file? My (possibly flawed) understanding is that we would prefer to use unit testing for the individual processors and design this test to be simpler?

Correct, see my comment above.

For test_waveform_browser.py, this involves actually drawing waveforms. How does this work for automatic testing? Is it more or less just going to look for if any errors are thrown, or is this more meant for someone to run on their own to visually confirm that waveforms are drawn correctly?

Yes, It is more or less just looking for if any errors are thrown. Testing image outputs is fairly complicated, obviously, and I'm not sure we'll ever have time to work on this...

gipert:
@jasondet the Sphinx bug has been fixed: https://github.com/sphinx-doc/sphinx/issues/10701

I'm going to merge this, I'll continue working on tests in future PRs.

PR #314, opened 2022-07-11
`raw.orca` cleanup

gipert:
As part of the refactor, we should make sure that only tested, standard-compliant code is part of the codebase. Since I don't think decoders for data takers (handled by ORCA) other than FlashCam will be refactored soon, I'm proposing to:

* remove `orca_digitizers.py` and `orca_pollers.py`
* remove `ORFlashCamListenerStatusDecoder` or raise a `NotImplementedError` like now.

These not-yet-refactored pieces of code should be stored elsewhere, in the meanwhile.


jasondet:
this should be on main not refactor, right?

jasondet (approved):

PR #318, opened 2022-07-15
fix else case for table/waveform reading

jasondet:


gipert:
We should add a unit test for this

PR #324, opened 2022-07-24
DSP Improvements

iguinn:
- Made debug output for processing_chain to not output values in arrays (this was TMI)
- lgdo Scalar writing in overwrite mode didn't work
- Changed behavior of the chan_config option, which lets you pass a dict pointing from channel name to dsp config files:
    - Before, you were required to provide a file for every channel in a file, and build_dsp would try to build every channel unless you use the lh5_tables option as well
    - Now, it loops through the dict and only builds provided channels; if it doesn't find a channel, it will skip it
    - Before you had to give the channel in chan_config as `chan#/raw`; now it will work with just `chan#`
- Fixed progress bar bug that was causing it to misreport the number of events in a file when it completed
- For non-integer pickoff times, instead of raising an exception, interpolate linearly between samples. This is important if we want to use pre-summing


src/pygama/dsp/_processors/fixed_time_pickoff.py
```diff
@@ -38,10 +39,9 @@ def fixed_time_pickoff(w_in, t_in, a_out):
...
-    if np.floor(t_in) != t_in:
-        raise DSPFatal('The pick-off index must be an integer')
-
     if int(t_in) < 0 or int(t_in) >= len(w_in):
         return
 
-    a_out[0] = w_in[int(t_in)]
+    i = int(t_in)
+    w = t_in-i
+    a_out[0] = (1-2)*w_in[i] + w*w_in[i+1]
```
jasondet, line 47: What is `(1-2)` doing there? Should it be `(1-w)`?

jasondet: also, I think it would be clearer to rename t_in as i_in since it is an index, not a time.

src/pygama/lgdo/lh5_store.py
```diff
@@ -622,8 +622,9 @@ def write_object(self,
 
         # scalars
         elif isinstance(obj, Scalar):
-            if name in group:
-                log.debug(f"overwriting '{name}' in '{group}'")
+            if wo_mode == 'o' and name in group:
```
jasondet, line 625: I also feel like if `name` is in `group` and `wo_mode` is `'w'` then we should emit an error here. @iguinn do you agree?

Also `wo_mode = 'a' should be okay for overwriting scalars, according to the `write_object` docstring.

jasondet:
Finally @iguinn can you check why the `Build documentation` test is failing? I couldn't figure it out from the [Details](https://github.com/legend-exp/pygama/runs/7489550632?check_suite_focus=true)

src/pygama/lgdo/lh5_store.py
jasondet: I made this change.

mmatteo:
@iguinn, I don't like the idea of changing the behaviour of the processor and like to have it as it is now. I would rather implement a specific processor that supports different types of interpolation and can replace fixed_time_pickoff if the user wants so

jasondet:
As of Ian’s PR, the code does not change behavior based on the cast of an input parameter, it does an interpolation every time. To my mind, this is the desired behavior of this processor. It gives consistent output independent of pre-summing, for example.

iguinn:
Ok, I"ve reverted the changes to fixed_time_pickoff. I'll create an issue for it, but until we fix it we shouldn't start using presumming

src/pygama/lgdo/lh5_store.py
iguinn: Ok, sounds good. I guess the other case is 'of' but in that case I think no check is needed since the file's overwritten so it shouldn't be possible to run into this? Either way, raising an error ourselves is a good idea, as before it was raising a somewhat confusing h5py error.

iguinn:
> Finally @iguinn can you check why the `Build documentation` test is failing? I couldn't figure it out from the [Details](https://github.com/legend-exp/pygama/runs/7489550632?check_suite_focus=true)

I can't tell either, the error message is pretty unhelpful here...It looks to me like other pull requests are running into the same issue, though, so it's not clear that changes here are the cause. If this PR is actually the cause, I did make changes to a function signature (specifically an annotation) on line 26 of build_dsp.py, so it's possible that could have caused it?

jasondet (approved):

jasondet:
The build_documentation failure is a [bug in sphinx](https://github.com/sphinx-doc/sphinx/issues/10701). I'm going to turn off this requirement for merging and merge.

PR #326, opened 2022-07-26
adding orca_packet test

jasondet:
See Issue https://github.com/legend-exp/pygama/issues/327


jasondet:
Again ignoring build_documentation failure due to [sphinx bug](https://github.com/sphinx-doc/sphinx/issues/10701).

gipert:
Could you please open an issue about this sphinx bug so I can remember to fix it? I guess we'll have to set a version bound in `setup.cfg` to exclude the buggy version.

PR #337, opened 2022-08-03
Data loading utilities, first implementation

gracesong312:
Still a lot of work to be done, but should be in a state that is usable for basic analysis, retrieving waveforms, etc.


src/pygama/flow/data_loader.py
```diff
@@ -0,0 +1,1060 @@
...
+from pygama.flow.fileDB import FileDB
+from pygama.lgdo import (
+    Array,
+    ArrayOfEqualSizedArrays,
+    LH5Store,
+    Struct,
+    Table,
+    WaveformTable,
+)
+from pygama.lgdo.VectorOfVectors import build_cl, explode_arrays, explode_cl
```
gipert, line 20: This cannot work...
```suggestion
from pygama.lgdo.vectorofvectors import build_cl, explode_arrays, explode_cl
```

src/pygama/flow/data_loader.py
```diff
@@ -0,0 +1,1060 @@
...
+                        tcm_tb = tcm_tables[0]
+                tcm_path = (
+                    self.data_dir
+                    + self.fileDB.tier_dirs[tcm_tier]
+                    + "/"
+                    + self.fileDB.df.iloc[file][f"{tcm_tier}_file"]
+                )
+                if not os.path.exists(tcm_path):
+                    raise FileNotFoundError(f"Can't find TCM file for {tcm_level}")
+                    return
```
gipert, line 498: Raising the exception is already enough
```suggestion

```

src/pygama/flow/data_loader.py
```diff
@@ -0,0 +1,1060 @@
...
+                                                    tcm_idx
+                                                ] = wf_table["values"].nda
+                                            else:  # wf_values is a VectorOfVectors
+                                                log.warning(
+                                                    f"Not sure how to handle waveforms with values of type {type(tier_table[col]['values'])} yet"
+                                                )
+                                        else:
+                                            log.warning(
+                                                f"Not sure how to handle column {col} of type {type(tier_table[col])} yet"
+                                            )
```
gipert, line 837: Wouldn't be better to throw `NotImplementedError`s here? Or is the execution safe to continue?

src/pygama/flow/data_loader.py
```diff
@@ -0,0 +1,1060 @@
...
+                    return load_out
+                else:
+                    log.warning(
+                        "I don't know how to output "
+                        + self.output_format
+                        + ", here is a lgdo.Table"
+                    )
+                    return load_out
+
+    def load_detector(self, det_id):  # TODO
```
gipert, line 948: I would suggest to throw `NotImplementedError`s in not-yet-implemented methods.

src/pygama/flow/data_loader.py
```diff
@@ -0,0 +1,1060 @@
...
+
+    cols = ["daqenergy", "trapEmax", "channel", "waveform"]
+    dl.set_output(fmt="lgdo.Table", merge_files=False, columns=cols)
+    # dl.set_cuts({"hit": "daqenergy > 100"})
+
+    # hit_cut_el = dl.gen_entry_list(save_output_columns=True)
+
+    tcm_el = dl.gen_entry_list(tcm_level="tcm", save_output_columns=True)
+
+    lout = dl.load(tcm_el, in_mem=True, rows="hit", tcm_level="tcm")
```
gipert, line 1060: We'll remove this part later. It would be really nice if you could refactor this tiny tutorial as an "examples" section in the `DataLoader` class docstring:
```
class DataLoader:
    """One-line description.
    
    Longer description.
    
    Examples
    --------
    >>> from pygama.flow import DataLoader
    >>> dl = DataLoader()
    >>> ...
    ```

File: src/pygama/flow/data_loader.py (hunk 3 above)
Comment (gracesong312) [reply src/pygama/flow/data_loader.py:837]: I was thinking it should be fine to load the other requested columns and just skip any that are vov

Comment (jasondet) [reply src/pygama/flow/data_loader.py:865]: I agree with @gracesong312 that it should be okay to continue with a warning. In this case we should add something like "Skipping..." to the end of the warning messages so that it's clear to the user that they won't have that data.

**Comment by jasondet** [comment]
@gracesong312 can you add a minimal test of the data_loader functionality? @gipert is there a set of e.g. raw+dsp files in the test suite she can use?

**Comment by gipert** [comment]
Yes, there is! But maybe requires some adjustments. Why don't we briefly discuss this on the phone @gracesong312?

File: src/pygama/flow/data_loader.py [hunk 6]
```diff
@@ -0,0 +1,1092 @@
...
+                # Convert f_entries DataFrame to Struct
+                f_dict = f_entries.to_dict("list")
+                f_struct = Struct(f_dict)
+                sto.write_object(f_struct, f"entries/{file}", f_output, wo_mode="o")
+        if in_mem:
+            return entries
+
+    def load(
+        self,
+        entry_list: pd.DataFrame = None,
```
gipert, line 699: You declare this argument `None` here but then raise an exception later if it's `None`. If the input is really needed then it should not be optional.

More generally, why isn't `gen_entry_list()` automatically called by `load()`, if the user does not want to add extra customization?

I also think that we should make the usage of the TCM the default, when required an event-oriented output (right now one is required to supply `tcm_level`). Most users will not want to think about this.

src/pygama/flow/data_loader.py
```diff
@@ -0,0 +1,1092 @@
...
+        in_mem: bool = False,
+        f_output: str = None,
+        rows: str = "hit",
+        tcm_level: str = None,
```
gipert, line 703: Maybe more self-documenting?
```suggestion
        in_memory: bool = False,
        output_file: str = None,
        orientation: str = "hit",
        tcm_level: str = None,
```

gipert (changes requested):
I've left a first round of comments, mainly about the interface. I've been trying to run this on the LNGS cluster (pwd is `/data1/users/pertoldi`):
```python
import numpy as np

from pygama.flow import DataLoader

dl = DataLoader(
    config="loader_config.json",
    filedb_config="fileDB_config.json"
)
dl.set_files("file_status == 26 and timestamp == '20220716T130443Z'")
dl.set_datastreams(np.arange(40, 45), "ch")
tcm_el = dl.gen_entry_list(tcm_level="tcm", save_output_columns=True)
lout = dl.load(tcm_el, in_mem=True, rows="hit", tcm_level="tcm")
````
But it fails with:
```
Traceback (most recent call last):
  File "/opt/anaconda3/lib/python3.9/site-packages/pandas/core/indexes/base.py", line 3621, in get_loc
    return self._engine.get_loc(casted_key)
  File "pandas/_libs/index.pyx", line 136, in pandas._libs.index.IndexEngine.get_loc
  File "pandas/_libs/index.pyx", line 163, in pandas._libs.index.IndexEngine.get_loc
  File "pandas/_libs/hashtable_class_helper.pxi", line 5198, in pandas._libs.hashtable.PyObjectHashTable.get_item
  File "pandas/_libs/hashtable_class_helper.pxi", line 5206, in pandas._libs.hashtable.PyObjectHashTable.get_item
KeyError: 'tcm_tables'

The above exception was the direct cause of the following exception:

Traceback (most recent call last):
  File "/data1/users/pertoldi/test.py", line 45, in <module>
    tcm_el = dl.gen_entry_list(tcm_level="tcm", save_output_columns=True)
  File "/data1/users/pertoldi/sw/src/legend/pygama/src/pygama/flow/data_loader.py", line 481, in gen_entry_list
    tcm_tables = self.fileDB.df.loc[file, f"{tcm_tier}_tables"]
  File "/opt/anaconda3/lib/python3.9/site-packages/pandas/core/indexing.py", line 960, in __getitem__
    return self.obj._get_value(*key, takeable=self._takeable)
  File "/opt/anaconda3/lib/python3.9/site-packages/pandas/core/frame.py", line 3615, in _get_value
    series = self._get_item_cache(col)
  File "/opt/anaconda3/lib/python3.9/site-packages/pandas/core/frame.py", line 3931, in _get_item_cache
    loc = self.columns.get_loc(item)
  File "/opt/anaconda3/lib/python3.9/site-packages/pandas/core/indexes/base.py", line 3623, in get_loc
    raise KeyError(key) from err
KeyError: 'tcm_tables'
```

File: src/pygama/flow/data_loader.py [hunk 8]
```diff
@@ -0,0 +1,1092 @@
...
+            log.info(self.fileDB.df.loc[self.file_list, columns])
+
+    def show_filedb(self, columns=None):
+        """
+        Wrapper to log the loaded FileDB
+        """
+        self.fileDB.show(columns)
+
+    def get_tiers_for_col(
+        self, columns: list | np.ndarray, merge_files: bool = None
```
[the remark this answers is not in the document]
gracesong312: If it's not told to do anything differently, it uses self.merge_files, which is why it defaults to None

src/pygama/flow/fileDB.py
```diff
@@ -0,0 +1,358 @@
...
+import numpy as np
+import pandas as pd
+from parse import parse
+
+from pygama.lgdo import Array, LH5Store, VectorOfVectors
+
+log = logging.getLogger(__name__)
+
+
+class FileDB:
```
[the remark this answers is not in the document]
gracesong312: Yes, we initially wanted to do that, but DataFrame is weirdly difficult to inherit from. https://pandas.pydata.org/pandas-docs/stable/development/extending.html#extending-subclassing-pandas 

The `self.fileDB.df.query()` syntax isn't ideal, but Jason recommended we just do that instead of figuring out their syntax just to define some extra fields.

src/pygama/flow/data_loader.py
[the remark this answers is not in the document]
gipert: Right, sorry

src/pygama/flow/file_db.py
```diff
@@ -0,0 +1,390 @@
...
+        """
+        log.debug(f"Writing database to {filename}")
+
+        sto = LH5Store()
+        sto.write_object(
+            Scalar(json.dumps(self.config)), "config", filename, wo_mode=wo_mode
+        )
+
+        if wo_mode in ["write_safe", "w", "overwrite_file", "of"]:
+            wo_mode = "a"
```
gracesong312, line 324: Why is the writing mode being changed to append? 

Also, after this in the code, `"a"` is just used directly regardless of the value of `wo_mode`

gipert: The idea is that if the file is created or overwritten in the first call to `write_object` a few lines above, all the rest must be appended. Other write modes that assume an existing file can be just passed to `write_object`. Does it make sense?

Now let me actually ass `wo_mode` in the subsequent calls, that's my fault.

gipert:
@jasondet @mmatteo @gracesong312 this looks mergeable to me. Do you have time to discuss about the status and to-do items on the phone in these days?

Pasting here my to-do list:
- Option to make a `DataLoader` without `FileDB`? Or produce it faster?
- Rename `FileDB.get_column_names()` and `FileDB.scan_files()` to something more consistent
- Implement CWD substitution in `data_dir` path from `DataLoader` / `FileDB` config (see dataflow)
- What is the possible max size of a FileDB object? should we implement chunked reading/writing?
- We need a convention for our JSON keywords
- If `in_memory=True`, also save loaded data as `DataLoader` attribute. Reset it (`del`?) in `DataLoader.reset()`.
- Implement `orientation="evt"`, awkward array output (maybe interesting for querying)?
- Implement `merge_files=True`

gipert:
I cannot load hit-tier data, just pushed a test to show the problem. Not sure what is going wrong. @gracesong312 could you please have a look?

gracesong312:
> I cannot load hit-tier data, just pushed a test to show the problem. Not sure what is going wrong. @gracesong312 could you please have a look?

Should be fixed, just a typo in the filedb-config

gipert:
dang... did not think about the config!

jasondet (approved):
This code looks quite good to me. I vote to merge and have people start to try to use it.

gipert:
Here are my old notes:
- Option to make a `DataLoader` without `FileDB`? Or produce it faster? If we want to have it automatically produced by the (automatic) data production, I guess it should be done incrementally. Think about the requirements for data monitoring
- Implement CWD substitution in `data_dir` path from `DataLoader` / `FileDB` config (see special `$_/` syntax in dataflow)
- What is the possible max size of a FileDB object? should we implement chunked reading/writing?
- Implement `orientation="evt"`, awkward array output (maybe interesting for querying)?
- Implement `merge_files=True`
- Check if can load a column from raw and one from hit. Say something to the user if trying to load from more than two levels
- Document code better, try to make it more readable

These are (also interface-related) items that might be better to fix before merging:
- We need a convention for our JSON keywords (dash, underscore, case...)
- Bypass generation of entry list if no cuts applied, saves quite some time!
- Rename `FileDB.get_column_names()` and `FileDB.scan_files()` to something more consistent (`scan_columns()` and `scan_files()`?)

jasondet:
Thanks @gipert!

- DataLoader can't really work without a FileDB. But the "slow" step is getting the table / column names out of each file. We can skip this by allowing the user to specify those in the config and just assume they are always present / the same. Then it's a small price to make a new FileDB. We'll start like this.
- @gipert can you describe a use case where CWD is desired?
- The FileDB size is dominated by (#tiers)*(filename_length) + (#channels)*int. For #tiers ~ 8, filename_length~32, and #channels~256, that's ~1k per file. So we can have 1M files before we reach a GB. I think it's safe to assume that no chunking should be necessary any time soon.
- We propose to implement `evt` output in the next step
- We will implement `merge_files` and make `merge_files=True` the default behavior
- The tests already include loading from raw + dsp + hit levels, see e.g. [this line](https://github.com/legend-exp/pygama/blob/980f750a4a4c6bc05487b7e76eceb6030c259e4f/tests/flow/test_data_loader.py#L41)
- Most functions etc have proper doc strings, and there are good code comments throughout, but we will do a sweep to make improvements. @gipert if you have any specific suggestions please let us know.

Interface-related items:
- The json here looks okay to me, is there something in particular that caught your attention? The broader context is important but I think we should pursue that outside of the context of this PR.
- Will implement bypassing of entry list if no cuts are applied.
- All functions that scan files to pull info into the FileDB will use the word "scan" in their name.

@gracesong312 post a comment here when these are all handled.

gipert:
> can you describe a use case where CWD is desired?

If the config is distributed together with the data (we could do this for the official LEGEND data), it would be nice to keep the whole thing relocatable. One could move the data (e.g. from LNGS to NERCS) and the loading would still work.

> Most functions etc have proper doc strings, and there are good code comments throughout, but we will do a sweep to make improvements. If you have any specific suggestions please let us know.

I found the code sometimes hard to understand. I don't have specific suggestions but in general I would recommend to 1) be generous with comments 2) reuse code by implementing helper functions, where possible 3) try to keep code nesting to a minimum (for loops and ifs). For example: in L309 of `data_loader.py` the `else` is unnecessary and the following code can be unindented.

> The json here looks okay to me, is there something in particular that caught your attention? The broader context is important but I think we should pursue that outside of the context of this PR.

It's OK right now because they are consistent, it was just a general comment. We should all agree on some general conventions (like using `-` or `_` to separate words in keys) for metadata soon

> All functions that scan files to pull info into the FileDB will use the word "scan" in their name.

Like. I also personally like more `build_entrylist()` rather then `gen_entry_list()` (and renaming similarly `gen_hit_entries()`), but it's a personal preference. We could maybe also prepend names of functions that the user is not supposed to use directly with an underscore?

jasondet:
Thanks @gipert.

- CWD: I think I understand the use case. I think we don't necessarily want CWD, which is the directory that the FileDB generation is being run from, but rather, we want to be able to specify data_dir relative to the location of the FileDB config file, right? In this case, I think $_ is not the right tool here; what is easier would be to interpret data_dir as being relative to the location of the config file. We can also implement environment variable expansion so that if the user really wanted CWD they can put $PWD in for data_dir. Sound okay?
- @gracesong312 will generally go through the code and try to do some cleanup along the lines of your other suggestions, although we didn't understand your example -- line 309 in data_loader.py is a comment

gipert:
>CWD: I think I understand the use case. I think we don't necessarily want CWD, which is the directory that the FileDB generation is being run from, but rather, we want to be able to specify data_dir relative to the location of the FileDB config file, right? In this case, I think $_ is not the right tool here; what is easier would be to interpret data_dir as being relative to the location of the config file (when it doesn't start with /). We can also implement environment variable expansion so that if the user really wanted CWD they can put $PWD in for data_dir. Sound okay?

Sounds great!

> although we didn't understand your example -- line 309 in data_loader.py is a comment

Sorry, L509!

gracesong312:
I made some of the changes we talked about but I am still working on merge_files, and I have some concerns about bypassing the `build_entry_list` part if we're not applying cuts.

The main problem is that without an `entry_list`, `load()` doesn't know how to store data from multiple levels. You can't ask for, say, `trapEmax_cal` and `ge_mult` at the same time without a TCM telling you which row to put each part of the data. 

If there are no cuts, `build_entry_list()` just loads a TCM and converts it to a dataframe. Then `load()` goes in and opens the hit/evt files, placing that data where the TCM tells it to. The hit/evt files are not being opened multiple times, so even if there is a way to get around calling `build_entry_list()` explicitly, I'm not sure that it's really more efficient.

The only case where it really is more efficient is if we aren't using a TCM, i.e. all the requested data comes from the hit-level. In which case, I could modify `build_hit_entries()` to grab all `output_columns` if there are no cuts, but I guess I have a conceptual issue with making a function that's supposed to get the entry list do the work of multiple functions. I'm not sure what's the best way to structure the code to handle this case. Thoughts?

gipert:
@gracesong312 @jasondet let's try to have this in v1.1.0, which I'd like to tag before the meeting.

About bypassing `build_entry_list()`: I would propose for now to just avoid calling [this](https://github.com/legend-exp/pygama/blob/058cd762003f14ab85a62f325970467f32729f0c/src/pygama/flow/data_loader.py#L718) if cuts are not set. Like for example setting `f"{low_level}_idx"` to `-1`? Can you make a proposal @gracesong312?

Then I would add what's missing in the docs (see comments from my last review above) and merge.

For the meeting: I think it would be nice if @gracesong312 could show a scheme of the FileDB + DataLoader workflow, also with disk I/O calls. It would be helpful to understand the performance of the workflow -- in the end, it's quite complicated.

PR #340, opened 2022-08-09
First implementation of `build_hit`

gipert:
Here's a first embarassingly simple implementation and here's some context:

* added `hit.build_hit` (not tested, I'll add tests later)
* evaluating column operations is something we'd surely would like to use in different context (e.g. `build_evt`). I propose the definition of `lgdo.Table.eval`, which does the job.
* if we want this routine to be efficient, I guess we could try to vectorize column operations with Numba. This has a non trivial implementation (most painful part is perhaps string expression parsing).
* before going for the point above, I though about making profit of [`pandas.DataFrame.eval`](https://pandas.pydata.org/pandas-docs/stable/reference/api/pandas.DataFrame.eval.html), which does exactly what we need. [Here](https://pandas.pydata.org/pandas-docs/stable/user_guide/enhancingperf.html#enhancingperf-eval) are some performance notes. I forced `numexpr` as a backend, which should be advantageous with tables with number of rows > 10⁴. I'm not sure this is optimal for us.
* one first downside is that we have to adopt Pandas's expressions syntax 
* another downside of this approach I'm really not comfortable with: lack of propagation of physical units. One idea would be to use [pint-pandas](https://github.com/hgrecco/pint-pandas), which seems to offer the desired functionality. I'm not sure the tool is stable enough though.

This would be the syntax for the `build_hit` JSON config files:
```js
{
    "O1": {  // name of the output column, as in DSP config files
        "expression": "@p1 + @p2 * a**2",  // pandas wants a "@" labeling external variables
        "parameters": {
            "p1": "2",
            "p2": "3"
        }
    },
    "O2": {
        "expression": "O1 - b"
    }
    // ...
}
```

We could test this on data and evaluate the performance, before deciding on how to proceed.

### DIFF
--- a/src/pygama/lgdo/table.py
+++ b/src/pygama/lgdo/table.py
@@ -95,8 +95,8 @@ class Table(Struct):
             elif len(obj) != new_size:
                 if do_warn:
                     log.warning(
-                        f"warning: resizing field {field}",
-                        f"with size {len(obj)} != {new_size}",
+                        f"warning: resizing field {field}"
+                        f"with size {len(obj)} != {new_size}"
                     )
                 if isinstance(obj, Table):
                     obj.resize(new_size)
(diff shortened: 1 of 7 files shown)


src/pygama/lgdo/table.py
```diff
@@ -205,3 +205,63 @@ def get_dataframe(self, cols: list[str] = None, copy: bool = False) -> pd.DataFr
...
+            - ``parameters`` is a dictionary of function parameters. Passed to
+              :meth:`pandas.DataFrame.eval` as `local_dict` argument.
+
+
+        Warning
+        -------
+        Blocks in `expr_config` must be ordered according to mutual dependency.
+        """
+        # don't want to modify input
+        df = self.get_dataframe().copy()
```
jasondet, line 253: is the copy really necessary? if it is, we can put `copy=True` in the arg of `get_dataframe()` (avoids making 2 dataframes). But I think we want to avoid reallocation of the data arrays, and just have the df use the ones in the table. They should not get modified by eval. See also below.

src/pygama/lgdo/table.py
```diff
@@ -205,3 +205,63 @@ def get_dataframe(self, cols: list[str] = None, copy: bool = False) -> pd.DataFr
...
+        for out_var, spec in expr_config.items():
+            df.eval(
+                f"{out_var} = {spec['expression']}",
+                parser="pandas",
+                engine="numexpr",
+                local_dict=spec["parameters"] if "parameters" in spec else None,
+                inplace=True,
+            )
+
+            out_tbl.add_column(out_var, Array(df[out_var].to_numpy()))
```
jasondet, line 265: I was wondering why you didn't just do e.g. 
`out_var = df.eval(...).to_numpy()`
inside the `for` loop but I guess adding the column to the df allows you to have expressions that build on previous expressions (like in the test code). I think it would be good to add a comment in the code that it is structured this way for this purpose.

jasondet:
very nice. I wonder if it would be good to also have `array.eval()` in lgdo?

src/pygama/lgdo/table.py
gipert: That is a shallow copy right? Means that only keys and references are actually copied. If this is correct, should be quite inexpensive. I'm copying it because the function would otherwise modify the input table by adding new columns. Maybe we could do some garbage collection at the end of the function and delete the copy?

src/pygama/lgdo/table.py
gipert: Yes, that behavior is intended. I will add a comment.

gipert:
> I wonder if it would be good to also have `array.eval()` in lgdo?

What use case do you have in mind?

jasondet:
I thought I had one but now that I think about it, I don't really. Nevermind.

src/pygama/lgdo/table.py
jasondet: you would need to do `df.copy(deep=False)` to get a shallow copy

If the df is built with `copy=True` then it's just a few of the table columns. When eval adds new columns to the df the lgdo table knows nothing about them. It should remain unmodified.

gipert: I see what you mean now. I will add the `copy=True` argument to `get_dataframe`.

gipert:
How we deal with the units issue in the short term? `hit` data does not get any units attribute attached at the moment.

jasondet:
I think in the short term we ignore it -- we rely on the user to know the units of the data we are working with, and then to manually set the `units` attribute as needed.

gipert:
Ok! I'll give this pint-pandas a try.

PR #341, opened 2022-08-10
Initial versions and updates of pargen scripts

ggmarshall:
Initial versions of the a/e calibration and the energy optimisation. Along with updates to the energy calibration from work on the L60 data. Additionally there are some bugfixes I found while running on this data. 
All these scripts are running and have been further tuned using the L60 data.


gipert:
There are quite a bit of things that should be reviewed here, should we rather do it at later times when refactoring?

jasondet:
Yes this code is urgent and we are going to refactor it anyway, so I will go ahead and merge it.

gipert:
Uhm there's still something wrong @ggmarshall:

https://github.com/legend-exp/pygama/runs/7819124376?check_suite_focus=true#step:5:68

Can you have a look?

jasondet:
The error is weird. If you open a python session and do
```
import scipy
print(scipy.__version__)
```
You will get something sensible like `1.8.1`. But if you put these two lines anywhere in our code and run `make` in docs, during `sphinx-build` it outputs `scipy.__version__`.

gipert:
I did this:

https://github.com/legend-exp/pygama/pull/341/files#diff-dc93db753c168ea353e31eddde92b9d47c704055ddc4d2f02286e052b781d1ccR780-R782

as suggested by @ggmarshall, is it ok?

ggmarshall:
Hi, this should give the same behaviour for the moment

docs/source/conf.py
```diff
@@ -46,6 +46,7 @@
     "matplotlib",
     "mplhep",
     "scipy",
+    "sklearn",
```
jasondet, line 49: @gipert this should be `scikit-learn` not `sklearn`, no?

jasondet: oh it looks like this is module name not the pip-install argument. This is indeed what makes build docs work, as long as `kernel=None` below.

jasondet:
Wow this is a weird error, but I was at least able to isolate it as a problem with docs generation, not @ggmarshall's code. I opened an issue:
https://github.com/legend-exp/pygama/issues/344

@ggmarshall, this code still works with `kernel = None` in energy_optimisation.py:780 ???

ggmarshall:
Yep I have tested and it is the same as the kernel I specified

gipert:
Ok thanks, I'll think about #344. In the meanwhile I'll go ahead and squash-merge this.

PR #345, opened 2022-08-15
More advanced `build_hit` interface

gipert:
As required by data production. This is how the hit config JSON file would look like:
```json

{
  "outputs": ["calE", "AoE", "A_max"],
  "operations": {
    "twice_trap_e_max": {
      "expression": "2 * trapEmax"
    },
    "calE": {
      "expression": "sqrt(@a + @b * twice_trap_e_max**2)",
      "parameters": {
        "a": "1.23",
        "b": "42.69"
      }
    },
    "AoE": {
      "expression": "A_max/calE"
    }
  }
}
```

### DIFF
--- a/setup.cfg
+++ b/setup.cfg
@@ -7,6 +7,7 @@ url = https://github.com/legend-exp/pygama
 author = The LEGEND collaboration
 maintainer = The LEGEND collaboration
 license = Apache-2.0
+license_file = LICENSE
 license_files = LICENSE
 classifiers =
     Development Status :: 4 - Beta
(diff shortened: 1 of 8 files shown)


gipert:
@jasondet I implemented `Struct.remove_field()` (and `Table.remove_column()`) to be able to customize the content of the output table from `build_hit()`. Does the code I wrote make any sense?

gipert:
Also, I'm not entirely sure buffering makes sense when relying on `pandas.DataFrame.eval()`. Or, at least, we should think about a meaningful default value for `buffer_len`.

jasondet (approved):

PR #347, opened 2022-08-17
added compass decoding capabilities to build_raw

SamuelBorden:
Created a CompassStreamer class for build_raw to use to decode CoMPASS files. Because of the layout of CoMPASS files, a further option was added to pass a config file to track enabled channels and waveform length.

### DIFF
--- a/src/pygama/raw/build_raw.py
+++ b/src/pygama/raw/build_raw.py
@@ -28,7 +28,7 @@ def build_raw(
     buffer_size: int = 8192,
     n_max: int = np.inf,
     overwrite: bool = False,
-    config_file: str = None,
+    compass_config_file: str = None,
     **kwargs,
 ) -> None:
     """Convert data into LEGEND HDF5 raw-tier format.
@@ -68,10 +68,10 @@ def build_raw(
     overwrite
         sets whether to overwrite the output file(s) if it (they) already exist.
 
-    config_file
+    compass_config_file
         Specification of config file, used for decoding CoMPASS files
 
-        - if None, CompassDecoder will eat the first packet to determine waveform length
+        - if None, CompassDecoder will sacrifice the first packet to determine waveform length
         - if a str ending in ``.json``, interpreted as a filename containing
           json-shorthand for the output specification (see :mod:`.compass.compass_event_decoder`).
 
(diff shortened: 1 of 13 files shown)


src/pygama/raw/compass/compass_header_decoder.py
```diff
@@ -0,0 +1,117 @@
...
+
+    .. code-block:: json
+
+        {
+        "model_name": "DT5730",
+        "adc_bitcount": 14,
+        "sample_rate": 500e6,
+        "v_range": 2.0,
+        "num_enabled_channels": 1,
+        "wf_len": 1000,  # in samples
```
gipert, line 31: Sphinx complains here because, strictly-speaking, comments are not supported in JSON. To circumvent this I usually force it to Javascript:
```rst
.. code.block:: js

    {
        "key": "value"  // comment
    }
```
Renders as
```js
{
    "key": "value"  // comment
}
```

src/pygama/raw/build_raw.py
```diff
@@ -66,6 +68,13 @@ def build_raw(
     overwrite
         sets whether to overwrite the output file(s) if it (they) already exist.
 
+    config_file
+        Specification of config file, used for decoding CoMPASS files
+
+        - if None, CompassDecoder will eat the first packet to determine waveform length
+        - if a str ending in ``.json``, interpreted as a filename containing
+          json-shorthand for the output specification (see :mod:`.compass.compass_event_decoder`).
+
```
gipert, line 77: Why not merging this extra config section in `out_spec`? Would avoid us having to add a new argument to `build_raw`.

gipert (changes requested):
The code looks really nice, thanks a lot for your work! I left a couple of comments.

If you want to make sure this decoder will not break in the future, I would recommend to upload a test file to [legend-testdata](https://github.com/legend-exp/legend-testdata) and implement some tests.

SamuelBorden:
I updated the hash in `/tests/conftest.py` to the most recent commit on the legend-testdata repo, I also added xmltodict as an install requirement because compass xml files cannot be parsed with plistlib

SamuelBorden:
Closing this, and then opening a new PR to main instead of the refactor branch.

PR #348, opened 2022-08-18
Disabled numba cache and add bound checks

mmatteo:
set the cache option to False. False it is the default value but I thought it was a good idea to leave it explicit. 
I also added enable the bounds-check and made uniform the guvectorize decorator throughout the repo.

@iguinn, the cache=true option keeps creating problems for the data production. Cache files are created under the user's home durectory (`~/.cache/numba`) the first time build_dsp is called. These files are not updated and the processing chain creates a seg fault if build_dsp is called with a different processor list. We can discuss more about this during a call.

Why did you enable the cache in the first place?

### DIFF
--- a/src/pygama/dsp/processing_chain.py
+++ b/src/pygama/dsp/processing_chain.py
@@ -1194,7 +1194,7 @@ class ProcessorManager:
 class UnitConversionManager(ProcessorManager):
     """A special processor manager for handling converting variables between unit systems."""
 
-    @vectorize(nopython=True, cache=True)
+    @vectorize(nopython=True, cache=False)
     def convert(buf_in, offset_in, offset_out, period_ratio):  # noqa: N805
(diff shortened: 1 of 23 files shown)


gipert:
>Why did you enable the cache in the first place?

It speeds up initialization from the second use on: https://numba.readthedocs.io/en/stable/user/jit.html#cache

> These files are not updated and the processing chain creates a seg fault if build_dsp is called with a different processor list

This is really strange. Why do you see this only in the data production?

Maybe you want to customize this? https://numba.readthedocs.io/en/stable/reference/envvars.html#envvar-NUMBA_CACHE_DIR. It could point to a temporary directory in the data-prod folder.

gipert:
After discussing on the phone, we decided to implement a set of default Numba options for our processors and give the user the possibility to dynamically change them at runtime. I will prepare a new PR.

PR #350, opened 2022-08-19
Bare-bones class for setting global Numba options

gipert:
Example usage:
```py
from pygama.dsp.utils import numba_defaults
from pygama.dsp import build_dsp

# must do this before explicitly importing pygama.dsp.processors!
numba_defaults.cache = False
numba_defaults.boundscheck = True

build_dsp(...) # processors imports happen here
```
* Compilation of procesors takes ~30 sec on my laptop. I think this is non-negligible and we should set `cache=True` by default.
* I would leave `boundscheck=False` default as in Numba.

To do:
- [ ] @iguinn what do we do with `nopython`? Do we really want to set it to `True`?
- [x] Update documentation

### DIFF
--- a/docs/source/manuals/build_dsp.rst
+++ b/docs/source/manuals/build_dsp.rst
@@ -45,7 +45,7 @@ Writing custom processors
     from numba import guvectorize
 
     from pygama.dsp.errors import DSPFatal
-
+    from pygama.dsp.utils import numba_defaults_kwargs as nb_kwargs
 
     # 2) Provide instructions to Numba
     #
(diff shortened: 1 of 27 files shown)


mmatteo (commented):
what happens in case kwargs options are in conflict with other given explicitely? Does the relative position of kwags and of the duplicated position affect the resolution of the conflict? If so, we should enforce a clear position for kwargs, e.g. last option)

mmatteo (approved):
what happens in case kwargs options are in conflict with other given explicitely? Does the relative position of kwags and of the duplicated position affect the resolution of the conflict? If so, we should enforce a clear position for kwargs, e.g. last option

gipert:
If a keyword argument is specified twice, Python will throw an exception. The only way to override a default option would be to write, for example:
```py
from pygama.dsp.utils import numba_defaults as nb_defaults

@guvectorize(["..."], "...", cache=nb_defaults.cache, boundscheck=True)
```
Not ideal, but I cannot think about anything smarter. The right way to manage all this would be Numba devs finally implementing a `NUMBA_CACHE` environment variable or similar. Hope they'll do that at some point so we can avoid all this complexity.

mmatteo:
okay, sounds good to me. As far as I'm concerned, it's ready to merge

mmatteo:
@gipert, I did not notice your proposal to set cache to true. I disagree with that, if numba people set it to false we should follow their lead and instruct developers to change it if needed. Common users should not find themself in trouble because of unintentional caching

Issue #319, opened 2022-07-18
`Processing_chain` : `lh5Store` `load_nda` fails if waveform table


patgo25:
Hi there,

again maybe its just me who is stupid, but:
`df = lh5.load_nda(file_raw, ['channel','waveform'],'spms/raw',idx_list=[*range(0,100)])`
results in error:
`AttributeError: 'WaveformTable' object has no attribute 'nda'`

But
`df = lh5.load_nda(file_raw, ['channel','waveform'],'spms/raw')`
~~works fine~~  also fails

for now I patchworked it with 97f2d80d09816f4ab77c7b355aee1604293b4c26

patgo25:
Also works now, thanks!

gipert:
The problem is that many LGDOs do not have a trivial representation as NumPy arrays (i.e. they don't have an `nda` attribute). Returning `values.nda` for a `WaveformTable` instance would work only in some cases (if `values` isn't a `VectorOfVectors`) and anyways ignores the `dt` and `t0` attributes.

The right way to fix this is to implement a `convert(library='numpy')` method in every LGDO class that returns a meaningful representation, then refactor `load_nda` to call this function on the read out object.

jasondet:
does it work if you do this?

`df = lh5.load_nda(file_raw, ['channel','waveform/values'],'spms/raw')`

patgo25:
Thank you, Jason. This works.
So it was me, in the end, being stupid.

jasondet:
not at all! glad we figured out an easier solution

PR #355, opened 2022-08-25
New DSP processors for SiPM waveforms

patgo25:
### Processors for SiPM processing

**New Processors:**
- `y_projection`: takes waveform, projects onto y-axis, and returns histogram in the form of bin content and bin edges
- `bl_from_y_projection`: takes the above-mentioned histogram and returns mean(=max of the histogram) and standard deviation
- `sipm_noise_regulation`: checks sipm signal candidates if they originate from noise by using the expected shape of the SiPM signal.
- `sum_at_positions`: takes indices and a waveform. returns sum of indicated waveform values
- `get_multi_time_pickoff`: Takes an array of indices and returns the corresponding array of values in a given waveform

**Modified Processor**: `get_multi_local_extrema`
- added absolute values an extremum must surpass to be tagged.

Added a "fix" for my issue  #319


patgo25:
pre-commit.ci autofix

gipert:
@SamuelBorden can you please have a look at `get_multi_local_extrema`?

gipert (changes requested):
My suggestions:

- changes in`get_multi_local_extrema.py` reintroduce docstrings before re-formatiing, please rebase on the current `main`
- I would merge `y_projection` and `bl_from_y_projection` into a single processor. Maybe call it something like `hypercurr_baseline_stats`?
- remove `sum_at_positions`
- remove `get_multi_time_pickoff`: we already have `multi_a_filter`

patgo25:
I disagree with merging the `y_projection` and `bl_from_y_projection` into a single processor. There is `bl_from_y_projection` which finds both mean and stddev and there is `bl_from_y_projection_at_mean` which takes a mean and returns the index to the mean and the stddev. Furthermore doing a projection has nothing to do with baseline reconstruction in the first place.  Maybe at some point, somebody will need a projection.

gipert (changes requested):
My opinionated list of requests:
- merge all histogram-related processors in a single file called `histogram.py`
- rename `y_projection()` to `histogram()`
- add a note to the docstring explaining why we don't just wrap `numpy.histogram()`
- rename `bl_from_y_projection()` to `histogram_stats()`
- make `histogram_stats()` compute all histogram-derived observables (remove `bl_from_y_projection_at_mean()`)
- use the correct wording for these observables: `mean` -> `mode`, `stddev` -> `fwhm`
- note: you compute the FWHM and then convert to variance, but this assumes Gauß distribution, which is not true in general. IMO the right thing to do is to return the FWHM. The user can then always convert FWHM to Gauß variance, if he thinks it applies

Summarizing, this is how the `histogram_stats` declaration could look like:
```py
def histogram_stats(
    weights_in: np.ndarray,
    edges_in: np.ndarray,
    mode_out: int,  # this is an index, the mode
    max_out: float,  # wf_in(mode_out)
    fwhm_out: float,
    assume_mode_zero : bool = False) -> np.ndarray:
    # do it
```

- include `from __future__ import annotations` at the top and add type hints everywhere
- update docstrings

patgo25:
To implement:

```
def histogram_stats(
    weights_in: np.ndarray,
    edges_in: np.ndarray,
    mode_out: int,  # this is an index, the mode
    max_out: float,  # wf_in(mode_out)
    fwhm_out: float,
    max_in : float) -> None:
```
I always need to pass a float default values are not possible. They conflict with the signature of the vectorizing decorator. . 

Two options 

1.  Have  two functions as before e.g 
    ```
    def histogram_get_max(
        weights_in: np.ndarray,
        edges_in: np.ndarray,
        mode_out: int,  # this is an index, the mode
        max_out: float,  # wf_in(mode_out)
        ) -> None:
    ```
    and
    ```
    def histogram_get_fwhm(
        weights_in: np.ndarray,
        edges_in: np.ndarray,
        max_in : float
        fwhm_out: float,) -> None:
    ```

2.  (Prefered by me) pass `np.nan` if the max should be searched automatically. For this parsing of NumPy constants is needed. I implemented one in pull request #365

iguinn:
I agree with Luigi's comments. In addition:
1) For `mode_out` above, do we want to pass an index or the bin center? Since we pass the FWHM in bin units rather than indices, I think we should do the same for the mode
2) For ```get_multi_local_extrema``` is this being used by anything outside of the sipm analysis? If so, I worry about changing the signature on others. I think Sam developed it specifically for sipms, so I think this is ok...
3) For `sipm_noise_regulation`:
  - I think we should rename this to something more generic, like `peak_snr_threshold` or something like that. This could potentially be useful for pileup detection in HPGe detectors or to look through FFTs or something like that
  - Can you please provide a sentence or two description of what it does in the doc string?
  - `idx_out` is going to be interspersed with NaNs. You should remove the NaNs and then pad it at the end
  - Also output the count of the number of peaks remaining. This will allow us to use `VectorOfVectors` to store this (once I actually finish implementing that :sweat_smile:, until then it will be a zero-padded array)
4) For `y_projection`/`histogram`, the argument `bin_in` is unnecessary. Since the guvectorize command requires being passed a pre-allocated size `(m)` array, we can use `len(proj_out)` instead. For the arguments passed, this would be:
```args = ["wf", "hist(100)", "edges"]``` instead of ```args = ["wf", 100, "hist(100)", "edges"]```

Note that if histogram used a fixed bin width, we could also have a lot of these use the ProcessingChain unit system, which would negate the need to pass around the array of bin edges. In this case you would do ```histogram(w_in, width_in, lowest_bin_out, hist_out)``` with ```args = ["wf", 5, "hist_offset", "hist(100, period=5, offset=hist_offset)"]```. Then, for the `histogram_stats`, you would output things in terms of indices and they would be automatically converted into y-units based on the period and offset. This would require a different behavior, though, since right now the output bin width can vary, and we'd need to handle the case where not everything fits into the histogram. Since we need to start running this soon, I think we should keep things as-is and consider changing this in the future.

**Review by iguinn — CHANGES_REQUESTED** [review]
I agree with Luigi's comments. In addition:
1) For `mode_out` above, do we want to pass an index or the bin center? Since we pass the FWHM in bin units rather than indices, I think we should do the same for the mode
2) For ```get_multi_local_extrema``` is this being used by anything outside of the sipm analysis? If so, I worry about changing the signature on others. I think Sam developed it specifically for sipms, so I think this is ok...
3) For `sipm_noise_regulation`:
  - I think we should rename this to something more generic, like `peak_snr_threshold` or something like that. This could potentially be useful for pileup detection in HPGe detectors or to look through FFTs or something like that
  - Can you please provide a sentence or two description of what it does in the doc string?
  - `idx_out` is going to be interspersed with NaNs. You should remove the NaNs and then pad it at the end
  - Also output the count of the number of peaks remaining. This will allow us to use `VectorOfVectors` to store this (once I actually finish implementing that :sweat_smile:, until then it will be a zero-padded array)
4) For `y_projection`/`histogram`, the argument `bin_in` is unnecessary. Since the guvectorize command requires being passed a pre-allocated size `(m)` array, we can use `len(proj_out)` instead. For the arguments passed, this would be:
```args = ["wf", "hist(100)", "edges"]``` instead of ```args = ["wf", 100, "hist(100)", "edges"]```

Note that if histogram used a fixed bin width, we could also have a lot of these use the ProcessingChain unit system, which would negate the need to pass around the array of bin edges. In this case you would do ```histogram(w_in, width_in, lowest_bin_out, hist_out)``` with ```args = ["wf", 5, "hist_offset", "hist(100, period=5, offset=hist_offset)"]```. Then, for the `histogram_stats`, you would output things in terms of indices and they would be automatically converted into y-units based on the period and offset. This would require a different behavior, though, since right now the output bin width can vary, and we'd need to handle the case where not everything fits into the histogram. Since we need to start running this soon, I think we should keep things as-is and consider changing this in the future.

**Comment by patgo25** [comment]
Thanks for the feedback, Ian.

> For mode_out above, do we want to pass an index or the bin center? Since we pass the FWHM in bin units rather than indices, I think we should do the same for the mode

For this is the `max_out: float,  # wf_in(mode_out)` field. But maybe in the future one needs the index rather than the bin center. hence I left it in, and leave it to the user which one to use. What do you think?

***
The other points I will fix first thing in the Morning
***

> Note that if histogram used a fixed bin width, we could also have a lot of these use the ProcessingChain unit system, which would negate the need to pass around the array of bin edges. In this case you would do histogram(w_in, width_in, lowest_bin_out, hist_out) with args = ["wf", 5, "hist_offset", "hist(100, period=5, offset=hist_offset)"]. Then, for the histogram_stats, you would output things in terms of indices and they would be automatically converted into y-units based on the period and offset. This would require a different behavior, though, since right now the output bin width can vary, and we'd need to handle the case where not everything fits into the histogram. Since we need to start running this soon, I think we should keep things as-is and consider changing this in the future.

Agreed, for now, this will do, and it is for sure not the final version ;)

**Review by iguinn — APPROVED** [review]

**Comment by iguinn** [comment]
Thanks Patrick, I'm satisfied with your changes.
I agree with you, that the index could potentially be used by other processors. We can think about how to get it to play nice with the ProcessingChain units system in the future; for now the important thing is having something that works!

**Comment by patgo25** [comment]
@gipert I added several tests for the DSP and HIT production of the SiPM test data in the last two commits. I think this concludes the features I wanted to add to this PR.

**Review by gipert — APPROVED** [review]

File: src/pygama/dsp/processors/histogram.py [hunk 1]
```diff
@@ -0,0 +1,145 @@
...
+
+from pygama.dsp.utils import numba_defaults_kwargs as nb_kwargs
+
+
+@guvectorize(
+    [
+        "void(float32[:], float32[:], float32[:])",
+        "void(float64[:], float64[:], float64[:])",
+    ],
+    "(n),(m),(m)",
```
gipert, line 16: @iguinn `borders_out` is actually one element longer than `weights_out`. How should the signature look like here?

src/pygama/lgdo/lh5_store.py
```diff
@@ -937,7 +937,10 @@ def load_nda(
                 data, _ = sto.read_object(f"{lh5_group}/{par}", f, idx=idx_list[ii])
             if not data:
                 continue
-            par_data[par].append(data.nda)
+            if isinstance(data, WaveformTable):  # HACK / FIXME
```
jasondet, line 940: Is this hack necessary? I thought one should be able to put 'waveform/values' in par_list to get this nda...

patgo25: #319

patgo25: Thanks @jasondet  Reverted the chang

patgo25:
To add the possibility of asking for a histogram with non-equally sized bins, I added changes in my dev rep
see here https://github.com/legend-exp/pygama/commit/d702a9729700dbc862248c2362c0c7c001cceef3 (@gipert )
if the given list just has one element (= the number of bins) an equal-sized histogram is created
if the given list is larger than one, the list is interpreted as bin widths for each bin.
(see relevant tests here: https://github.com/legend-exp/pygama/commit/64d4fe102987f598f0699f3b7acb096295533b9d

For this to work, a list must be able to be parsed from the JSON file. See the pull request for this here: #371 @iguinn  can you take a look?

If we still want to include this now, I can merge it here from my dev branch as soon as #371 is merged. Otherwise, we can add this feature also in a later PR since atm nobody needs a non-linear binned histogram (at the moment).

@jasondet @gipert @iguinn Let me know what you think

gipert:
My preference would be to merge this and discuss variable binning in a separate PR. Anyways, let's discuss at today's call.

patgo25:
pre-commit.ci autofix

gipert:
Thanks a lot @patgo25 for the work! 🎉

Issue #325, opened 2022-07-25
Interpolation modes in fixed_time_pickoff


iguinn:
For the fixed_time_pickoff processor, the current behavior is to raise an exception and halt processing if we run into a non-integer index. This particularly creates a problem if we are using presumming or down-sampling to speed up the trap-filter, but fully-sampled waveform for t0, in which case we will often run into cases where the timepoint falls between trap-samples. After discussing with @mmatteo and @jasondet we decided that the most sensible thing to do is add a flag for different interpolation modes.

Proposal:
Add an additional field for the interpolation mode. I think we were thinking either a string or a character (int is also an option but could be less clear). One mode would keep the current behavior, one would round the timepoint to the nearest integer, one would linearly interpolate, and we could add more in the future. My proposed options would be:
'i'/'index': current behavior
'r'/'round' (or 's'/'step'?) round the timepoint to the nearest integer; this would be a step-like interpolation
'l'/'linear': linear interpolation

Alternate proposal:
Add a boolean interpolate flag (if false use current behavior) with an integer for the order of interpolation (0 for step-like, 1 for linear, etc.). Or even just an integer, with -1 for current behavior.

Note on up-sampling:
Matteo also suggested using an up-sampler that would interpolate the whole waveform; however, for a single pickoff, this would be a lot slower than having the interpolation done by a single processor.

Possibility of renaming:
Jason suggested renaming to "fixed_sample_pickoff". We could consider leaving in place "fixed_time_pickoff" as a deprecated processor with the plan of removing it in the future and implementing the interpolation mode under the new name?

iguinn:
I have made an implementation of what we have discussed here: #360 
- `fixed_sample_pickoff` has the old functionality of raising exceptions for non-integer time inputs
- `fixed_time_pickoff` has an extra argument `mode_in` that accepts a character. Current options are:
  - 'i': require integer index (same as fixed_sample_pickoff)
  - 'n': nearest-neighbor; round to nearest integer index
  - 'l': linear interpolation
  - 'h': cubic hermite spline interpolation (continuous and once-differentiable; requires 4 nearest samples)
  - 's': cubic natural spline interpolation (continuous and twice-differentiable; requires full waveform)

Numba guvectorize does not make passing characters/strings easy. It requires treating them as an array of 8-bit integers. When calling `fixed_time_pickoff` directly, you need to convert the character string into a numpy array before passing it in like:
[...]

PR #360, opened 2022-09-02
Added interpolation option to `fixed_time_pickoff()` processor

iguinn:
Note: this pull request is not yet ready to accept, as changes to processing_chain are required to handle string inputs

- Renamed old `fixed_time_pickoff` to `fixed_sample_pickoff`
- `fixed_time_pickoff` now has a character argument to accept an interpolation mode. Options are:
  - 'i': require integer index (same as `fixed_sample_pickoff`)
  - 'n': nearest-neighbor; round to nearest integer index
  - 'l': linear interpolation
  - 'h': cubic hermite spline interpolation (continuous and once-differentiable; requires 4 nearest samples)
  - 's': cubic natural spline interpolation (continuous and twice-differentiable; requires full waveform)

See issue https://github.com/legend-exp/pygama/issues/325 for more discussion

### DIFF
--- a/src/pygama/dsp/processing_chain.py
+++ b/src/pygama/dsp/processing_chain.py
@@ -1150,6 +1150,16 @@ class ProcessorManager:
                         arshape.insert(len(arshape) + idim + 1, 1)
                 param = param.reshape(tuple(arshape))
 
+            elif isinstance(param, str):
+                # Convert string into integer buffer if appropriate
+                if np.issubdtype(dtype, np.integer):
+                    try:
+                        param = np.frombuffer(param.encode('ascii'), dtype).reshape(shape)
+                    except(ValueError):
+                        raise ProcessingChainError(
+                            f"could not convert string '{param}' into"
+                            f"byte-array of type {dtype} and shape {shape}")
+
             elif param is not None:
                 # Convert scalar to right type, including units
                 if isinstance(param, (Quantity, Unit)):
--- a/src/pygama/dsp/processing_chain.py
+++ b/src/pygama/dsp/processing_chain.py
@@ -1693,76 +1693,87 @@ def build_processing_chain(
             log.warning(
                 f"I don't know what to do with {input_par}. Building output without it!"
             )
-        proc_chain.link_input_buffer(input_par, buf_in)
+        try:
+            proc_chain.link_input_buffer(input_par, buf_in)
+        except Exception as e:
+            raise ProcessingChainError(
+                f"Exception raised while linking input buffer {input_par}."
+            ) from e
 
     # now add the processors
     for proc_par in proc_par_list:
         recipe = processors[proc_par]
-        module = importlib.import_module(recipe["module"])
-        func = getattr(module, recipe["function"])
-        args = recipe["args"]
-
-        # Initialize the new variables, if needed
-        if "unit" in recipe:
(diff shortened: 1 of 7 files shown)


iguinn:
The needed processing_chain changes for passing strings as numpy arrays of `int8`s have been added, so this pull request is ready now.
Additional changes from recent commits:
1) Added floor (`f`)/ceiling (`c`) interpolation modes
2) Added `interpolating_upsampler` which upsamples and then fills all the values from same set of interpolation modes used in `fixed_time_pickoff`
3) Added `interpolated_time_point_thresh` which seeks threshold crossings in the same way as `time_point_thresh`, but is capable of returning time points between samples, using the same set of interpolation modes as before (the `h` and `s` modes have not been implemented; they also present an interesting complication, since the splines can overshoot the samples and cross the threshold between different pairs of samples than `time_point_thresh`...)

jasondet (approved):
Looks good to me. We will need to make it clear that applications (like L60 processing) wanting fixed sample pickoffs need to change their dsp.json's

jasondet:
@ggmarshall or @mmatteo let us know if you are ready for us to merge this

gipert:
I don't understand why this flake8 error shows up now, does it make sense to you @jasondet?

jasondet:
It only just now showed up for the first time... it's a reasonable error though...

Let's see if it's still there after merging other PRs...

jasondet:
flake8 error is fixed in https://github.com/legend-exp/pygama/pull/373, will pull it in

jasondet:
I chatted with @mmatteo over slack. He pointed out that if fixed_time_pickoff() with option "i" is equivalent to fixed_sample_pickoff(), then maybe we shouldn't keep the latter. @iguinn what do you think? Since the signature of fixed_time_pickoff() has changed, no one can accidentally get the wrong behavior, so I think it might be okay not to have fixed_sample_pickoff().

iguinn:
One subtle point is that I added type signatures that accept integers instead of floats for `fixed_sample_pickoff`. This actually goes against our normal ethos of treating indices/times always as floats, and would only really matter for using things like `argmax` that pass integers anyway.

I don't have a problem with getting rid of it though.

jasondet:
@iguinn yes I think since the user is forced to insert an `'i'` to get fixed-sample behavior there is no extra safety from the explicit int typing in the gufunc directive. So I think we can remove fixed_sample_pickoff as suggested by @mmatteo after all.

when you do that, can you also fix the test failure? it looks like a numbification issue but I don't understand why it didn't show up before now.

iguinn:
I can't understand why the tests are failing. From the logs it looks like it is calling trying to call the processor without including the interpolation argument, but when I look at the configuration for the test the argument is there; I'm also able to compile the test file from the legend-testdata repo with the dsp config on my own. It's also failing on the sipm dsp test, which doesn't change at all with this PR.

jasondet:
@gipert can you take a look?

iguinn:
Oooooh found the problem. `multi_a_filter` was modified to call `fixed_time_pickoff` in a parallel branch, so once that got merged into this commit it caused problems. Now the tests seem to be passing.

gipert:
hope we'll have per-processor tests at some point to catch these more easily...

mmatteo:
catching up now. Everything sounds good

Issue #362, opened 2022-09-06
`Table.eval()`  converting table to dataframe makes it impossible to use with rows of e.g `ArrayOfEqualSizedArrays`


patgo25:
```py
df = self.get_dataframe()
# evaluate expressions one-by-one (in order) to make sure expression
# dependencies are satisfied
for out_var, spec in expr_config.items():
    df.eval(
        f"{out_var} = {spec['expression']}",
        parser="pandas",
        engine="numexpr",  # this should be faster than Python's native eval() for n_rows > 1E4, see Pandas docs
        local_dict=spec["parameters"] if "parameters" in spec else None,
        inplace=True,
    )
    # add column to output LGDO Table
    out_tbl.add_column(out_var, Array(df[out_var].to_numpy()))
```
Converts the table first into pandas dataframe, uses the pandas eval function and then converts it back to a table. This makes it impossible to use columns of e.g 'ArrayOfEqualSizedArrays'.

Why not use `numexpr `directly and apply it to the `nda` of the columns? 
My temporary fix see [here](https://github.com/patgo25/pygama/compare/cherry_picking...patgo25:pygama:dev)

PR #363, opened 2022-09-06
`lgdo.Table.eval()` change to use NumExpr2

patgo25:
Changed `LGDO::Table.eval()` from using pandas eval() function to using directly the NumExpr module.
For two reasons:
1.  `LGDO::Table` -> `pandas data frame' ->` LGDO::Table` seems inefficient to me
2. The "old" code can't evaluate `ArrayOfEqualSizedArrays` .

Further improvements:
- JSON syntax is more like DSP syntax: `expression:  "@a+@b*var" `--> `expression:"a+b*var"`

Changes to 'test_table_eval.py`:
- Included test form 'ArrayOfEqualSizedArrays' evaluations (`ArrayOfEqualSizedArrays `-> `ArrayOfEqualSizedArrays `and `ArrayOfEqualSizedArrays `-> `Array  `(i.e sum(var, axis=1)) manipulations.
- Extended tests to test logical operations

Change to `test_build_hit.py`:
- Adjusted the test JSON configs to the new format

Solves issue #362

### DIFF
--- a/src/pygama/lgdo/table.py
+++ b/src/pygama/lgdo/table.py
@@ -10,7 +10,11 @@ from typing import Any, Union
 import pandas as pd
 from pandas.io.formats import format as fmt
 
+import numexpr as ne
+
 from pygama.lgdo.array import Array
+from pygama.lgdo.arrayofequalsizedarrays import ArrayOfEqualSizedArrays
+from pygama.lgdo.fixedsizearray import FixedSizeArray
 from pygama.lgdo.scalar import Scalar
 from pygama.lgdo.struct import Struct
 from pygama.lgdo.vectorofvectors import VectorOfVectors
@@ -211,7 +215,7 @@ class Table(Struct):
 
         return df
 
-    def eval(self, expr_config: dict) -> Table:
+    def eval(self, expr_config: dict,usepdeval=True) -> Table:
         """Apply column operations to the table and return a new table holding
         the resulting columns.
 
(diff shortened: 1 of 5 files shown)


src/pygama/lgdo/table.py
```diff
@@ -243,33 +247,70 @@ def eval(self, expr_config: dict) -> Table:
...
+                    raise TypeError("Data of type VectorOfVectors not supported (yet)")
+
+            out_data = ne.evaluate(
+                f"{spec['expression']}",
+                local_dict=dict(in_vars, **spec["parameters"])
+                if "parameters" in spec
+                else in_vars,
+                global_dict=None,
+                optimization="moderate",  # Slow, but calculation is accurate (alternative "aggressive")
+                truediv="auto",
```
gipert, line 290: I would remove these (i.e. default behavior), I don't think we care. My intuition is that the default `optimization='aggressive'` is fine, we are not doing precision calculations (or non-trivial floating-point operations).

gipert (commented):
Like this very much, thanks a lot for the work @patgo25. Can you briefly report about this at the analysis call today? We can proceed with merging then.

src/pygama/lgdo/table.py
patgo25: Done
i.e optimization is now 'aggressive' (default option) --> faster but small errors can occur

gipert (approved):

jasondet:
I like this change. However I think the function signature should be a bit more apparent, right now all of the functionality is hidden in the various ways one can make an entry in expr_config. So one can't guess how to use it just by looking at the args. I would prefer something more like the following:

`tbl.eval(expr=None, pars_dict=None, out_name=None, append_to_self=False, return_nda=False, expr_dict=None)`

This matches more closely the signature of ne.evaluate(). If out_name is None you get an lgdo back (Array or ArrayOfEqualSizedArrays), unless return_nda=True in which case one just gets the ndarray returned by ne.evaluate. If out_name is not None, you get a table back with the column named by out_name; if append_to_self is true the column gets added to the table being evaluated. Then if expr_dict is not None you get something similar to the present behavior, and it will call itself recursively and build the table to return, or append to self according to the append_to_self option.

patgo25:
Hi Jason, I didn't touch the declaration of `tbl.eval()`, I am not sure where all in pygama `tbl.eval()` is referenced (which would need to be changed accordingly), and if I understood your envisioned changes correctly. If you have some time, I would propose you change this in a separate pull request since in my opinion, this is a more significant change beyond the scope of this PR.

jasondet:
That's a good point @patgo25. Let's proceed as you recommended. I'll sign off on it and merge it as-is.

jasondet (approved):

PR #364, opened 2022-09-08
Fixes to ProcessingChain

iguinn:
Fixed problems that were preventing unit conversion in specific cases. One example where this caused problems was using the `numpy.argmax` processor to identify maximum timepoint
- timepoint was not assigned a coordinate grid because the unitless axis variable was being used to try to deduce the unit
- When the coordinate is an integer type, it could not convert between different coordinate grids. Now if is_coord is set, when conversions will use `float64` to avoid this
- Replaced some `DOCME`s with words

### DIFF
--- a/src/pygama/dsp/processing_chain.py
+++ b/src/pygama/dsp/processing_chain.py
@@ -128,19 +128,22 @@ class ProcChainVar:
         Parameters
         ----------
         proc_chain
-            DOCME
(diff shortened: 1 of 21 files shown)


jasondet:
Looks good to me -- would be nice to add a test covering this case but I'm going to go ahead and merge this since it's needed soon.

PR #365, opened 2022-09-09
Adding `numpy.constants` parsing to `dsp.processing_chain`

patgo25:
This allows to use numpy constants (e.g.: `np.pi` , `np.e` , `np.inf`, `np.nan`) as arguments in the JSON file:

```
"calc1":{
            "function": "subtract",
            "module": "numpy",
            "args": ["foo", "5+2*(1+np.pi)*np.e", "calc1"]
        },
"calc2":{
            "function": "subtract",
            "module": "numpy",
            "args": ["foo", "np.inf", "calc2"]
}
"calc3":{
            "function": "subtract",
            "module": "numpy",
            "args": ["foo", "np.nan", "calc3"]
}
"calc4":{
            "function": "subtract",
            "module": "numpy",
            "args": ["foo", "5*np.pi*bar", "calc4"]
}
```
Added `tests\dsp\test_numpy_constants_parsing.py` and `tests\dsp\configs\numpy-parsing.json` for testing

Limitations:
```
"calcFAIL":{
            "function": "subtract",
            "module": "numpy",
            "args": ["foo", "np.inf*bar", "calcFAIL"]
        }
```
Raises exception since the sign of bar is not known. But who would do this anyhow?

This can be very useful for processors with default values e.g.

### DIFF
--- a/src/pygama/dsp/processing_chain.py
+++ b/src/pygama/dsp/processing_chain.py
@@ -15,7 +15,6 @@ from copy import deepcopy
 from dataclasses import dataclass
 from typing import Any, Union
 
-import numexpr as ne
 import numpy as np
 from numba import vectorize
 
@@ -818,6 +817,18 @@ class ProcessingChain:
 
         # for name.attribute
         elif isinstance(node, ast.Attribute):
+            # If we are looking for an attribute of a module (e.g. np.pi)
+            if node.value.id in self.module_list:
+                mod = self.module_list[node.value.id]
+                attr = getattr(mod, node.attr)
+                if not isinstance(attr, (int, float)):
+                    raise ProcessingChainError(
+                        f"Attribute {node.attr} from {node.value} is not"
+                        f"an int or float..."
+                    )
+                return attr
+
+            # Otherwise this is probably a ProcChainVar
             val = self._parse_expr(node.value, expr, dry_run, var_name_list)
             if val is None:
                 return None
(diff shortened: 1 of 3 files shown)


iguinn (changes requested):
Hey Patrick,
This is a good idea for a feature to have! However, I have tried to move away from using regexs for parsing, and have tried to put as much parsing as possible in `_parse_expr`. This uses python's built in parsing module which makes it very flexible and maintainable. I've made changes to do it this way, and submitted a PR to your fork so you try it out. The way I've done it will grab any int/float attributes from an modules added to a list, which right now just includes numpy.

Also, `numexpr` is not a default python module, so adding this would create a new dependency for pygama, which would require a bit more discussion/review, so I've gotten rid of this.

Ian

patgo25:
Thanks Ian.

I like your approach much more. 
The only divergence I see is that it doesn't support infinity arithmetic, but I guess nobody would do this anyhow ;)

I quickly change the tests and commit them here

patgo25:
@iguinn Numexpr is already anyhow needed for pygama in `lgdo.Table.eval()` since the the pandas eval function depends on it (see [here](https://pandas.pydata.org/docs/user_guide/enhancingperf.html#expression-evaluation-via-eval) ) 

Maybe you also want to take a look at #363 which avoids  using pandas.eval(), to support `ArrayOfEqualSizedArrays` (which currently isn't)

gipert:
Seems to me all issues have been addressed here. I'll proceed (squashing and) merging.

iguinn:
Looks good! Also agree that numexpr is fine to use; I didn't realize that it had been added as an installation prereq already.

PR #369, opened 2022-09-13
Data Cleaning Processors

ggmarshall:
I have added two processors for data cleaning. The first is a processor to invert the polarity for identifying discharge type events and the second is a time over threshold processor for looking at saturated events.


gipert:
Nice. Can you please also briefly document the processors in the docstrings?

ggmarshall:
done

src/pygama/dsp/processors/invert_polarity.py
```diff
@@ -0,0 +1,36 @@
...
+
+
+@guvectorize(
+    ["void(float32[:], float32[:])", "void(float64[:], float64[:])"],
+    "(n)->(n)",
+    **nb_kwargs,
+)
+def invert_polarity(w_in: np.ndarray, w_out: np.ndarray) -> None:
+    """Inverts the polarity of the waveform.
+    Parameters
```
gipert, line 16:
```suggestion
    """Inverts the polarity of the waveform.
    
    Parameters
```
The build will fail if you don't have spaces between sections

jasondet:
@ggmarshall is invert_polarity really necessary? Can't you get the same behavior using numpy.multiply with -1?

jasondet:
@gipert just wondering -- how does codecov report an increase in coverage in a PR with two new files and no new tests?

ggmarshall:
Good point, I have removed it

gipert:
> how does codecov report an increase in coverage in a PR with two new files and no new tests?

he saw your comment and changed his mind...

mmatteo:
Should the processor also recalibrate the adc scale such to have only positive values? Are all our processors able to handle negative adc values?

PR #371, opened 2022-09-14
Add possibility to parse a list to the parameter entry of the DSP JSON

patgo25:
Adding the possibility to parse a list to the parameter entry of the DSP JSON e.g:
```
...
"args": ["w_in","[1,2.5,3,4,1,6,10,8,2,1]","w_out"],
...
```

### DIFF
--- a/src/pygama/dsp/processing_chain.py
+++ b/src/pygama/dsp/processing_chain.py
@@ -9,6 +9,7 @@ import importlib
 import itertools as it
 import json
 import logging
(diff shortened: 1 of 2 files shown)


iguinn:
This could be a nice feature to have! Do you have a test case for it that we could add?

patgo25:
@iguinn added a simple test with numpy.add processor

gipert:
@iguinn can this be merged now?

PR #374, opened 2022-09-16
Improved handling of Numba default options, add `PYGAMA_*` shell variables

iguinn:
1. Instead of defaulting to no caching, default to no caching unless the `PYGAMA_CACHE` environment variable is set to `1`, `t` or `true`. Overriding at the start of the program is still possible
2. Changed the numba_defaults object to make it easier to change a single value from default for a specific processor. Instead of having to specifically invoke the default values in the `@guvectorize` decorator, now you can call `**nb_kwargs(cache=False)` to override that specific kwarg while using defaults in other cases.
The default values object now inherits from `MutableMapping` so that it acts like a dict itself; this enables us to still use `**nb_kwargs` as before.
4. For all (I think) of the processors that rely on factories, set caching to false regardless of the default. For these processors, the kernels get hard-compiled into the numba IR; if we cache these, then we will not be able to change the kernels!

### DIFF
--- a/docs/source/manuals/build_dsp.rst
+++ b/docs/source/manuals/build_dsp.rst
@@ -16,8 +16,9 @@ to learn about their meaning.
    Some special processors override default option values.
 
 The environment variables that can be set are:
-:PYGAMA_CACHE: set caching behavior (default False)
-:PYGAMA_BOUNDSCHECK: set automatic bounds checking (default False)
+
+:``PYGAMA_CACHE``: Set caching behavior (default false)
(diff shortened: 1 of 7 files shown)


gipert (approved):
Love this feature! Thanks a lot. Will be much easier with the env variable.

One last thing: could you  also update docs in the manual? https://github.com/legend-exp/pygama/blob/main/docs/source/manuals/build_dsp.rst#global-numba-options

gipert:
@mmatteo @ggmarshall anything against merging this? The PR does not change current behavior, just adds functionality.

PR #379, opened 2022-09-22
Updates to pargen scripts

ggmarshall:
Changes to new par_hit format without @ 
Improvements to A/E, in particular improved cut and sf determination.
Updated extract tau to be able to do double pole zeros also.
More robust cut determination.
Various upgrades to energy optimisation: faster event selection, better fitting etc.
Also changed the max value for the saturation processor.


gipert:
Codespell thinks you have misspelled `ect`, you can either add an exception [here](https://github.com/legend-exp/pygama/blob/main/.pre-commit-config.yaml#L92) or change the variable name

Issue #316, opened 2022-07-12
Plotting `numpy.ndarray` objects with `WaveformBrowser`


gipert:
The array-like output of some processors cannot be natively handled by the `WaveformBrowser` because of missing info about sampling period and offset (as instance `avg_current` or `upsampler`). A workaround is to explicitly set these values in the JSON spec, for example:
```json
"curr": {
    "function": "avg_current",
    "module": "pygama.dsp.processors",
    "args": ["waveform", 10, "curr(len(waveform)-10, 'f', period=waveform.period)"],
    "unit": "ADC/sample"
},
```
Maybe we can implement a mechanism to set default period and offset values, but we should at least document all this.

Issue #307, opened 2022-07-05
Make tutorials work on the refactored code


jasondet:

gipert:
And make them use public data from https://github.com/legend-exp/legend-testdata

PR #387, opened 2022-10-05
Fix `VectorOfVectors` read/write and add fancy indexing support

jasondet:

PR #394, opened 2022-10-20
Update `IntroToDSP.ipynb` tutorial to work with `pygama` 1.1.0

slwatkins:
I've updated @iguinn's `IntroToDSP.ipynb` tutorial to work with the current refactored version of `pygama` and added it to the notebooks folder in the sphinx documentation. I've built the docs on my local, and everything works well.

Most of the text is the same as Ian's version --- I updated the JSON config, the various `WaveformBrowser` kwargs being called, use the [legend-testdata-py](https://github.com/legend-exp/legend-testdata-py) package, and added the fix to the current calculation that's mentioned in #316 so the processors are plotted on the correct timescale.

Let me know if there's anything I should add to it!

I should note this is related to Issue #307.

----


gipert:
Thanks Sam for the work! Here are some suggestions from my side:
* remove "Originally written by [...]"
* remove `prereqs` from JSON DSP config
* set `plt.rcParams["figure.figsize"] = (14, 4); plt.rcParams["figure.facecolor"] = "white"; plt.rcParams["font.size"] = 12` like in the `WaveformBrowser` tutorial: https://github.com/legend-exp/pygama/blob/main/docs/source/notebooks/WaveformBrowser.ipynb
* use `us` units in `WaveformBrowser`'s `x_lim` argument
* I would remove the conclusions, they don't fit pygama's manual imo
* Maybe we want to conclude the tutorial by running `build_dsp()` on the test file and show the ouput file contents?
* Try to use more markdown markup where possible, will look nicer on the website!

slwatkins:
@gipert Thanks for the feedback! I've updated the tutorial with your suggestions, and added a bit at the end on using `build_dsp`. I also advertise the other tutorials that exist for more info, so that we don't have the same info in multiple spots.

gipert:
Thanks a lot! I just pushed some minor changes. One last thing: am I wrong or the last cell is supposed to print nicely on the dark theme only? If yes, I would just put
```py
load_dfs('./example_dsp_file.lh5', obj.keys(), 'geds/dsp')
```
in that cell.

gipert:
Can you also remove the old notebook in `tutorials/`?

FYI Black will run on Jupyter notebooks too, once #387 is merged

slwatkins:
@gipert Just removed the old tutorial!

For the styling I added - I saw that when I compiled the docs on my local, it was not very readable (default was dark grey text on a black background) on the documentation site. I'm assuming it'd be the same for the online doc site - I'm happy to remove those lines, but in that case maybe we should add a default Pandas printing style to the docs?

slwatkins:
Ah, nevermind I misunderstood what you were suggesting -> sounds good I'll just remove the printing of the DataFrame!

iguinn:
For reviewing purposes, what's the best way to see the output of a jupyter notebook? Is there a way to see it after it has been run, like this: https://pygama.readthedocs.io/en/stable/notebooks/LH5Files.html, or do I just want to run it myself?

Also, for the future, we should try to use `git mv` when moving files to preserve the history. I don't think it's a big deal for this one since it didn't have much of a history, but as things move around it's something to keep in mind.

gipert:
@iguinn if you click on "details" to the right of the readthedocs GitHub actions check it should forward you to the resulting HTML pages.

gipert:
@iguinn this PR looks good to me. Feel free to merge when you're done reviewing.

PR #399, opened 2022-10-21
Add `fixed_time_pickoff` unit test

slwatkins:
I wrote a simple unit test for all of the various cases of `fixed_time_pickoff`, and added it to the testing suite. I have a couple of questions before marking this as ready to review.

1. Are we okay with using the simple ndarrays for this test as I did? The outputs are more easily understand than using LEGEND test data, but perhaps there was a decision somewhere to always use test data.
2. This doesn't actually do anything to the code coverage, since the Python code itself isn't actually run due to the JIT compilation. Do we want to exclude these numba-wrapped functions from the coverage calculations? I'm not sure of a way of including them in the coverage other than creating two versions of each (like this: https://stackoverflow.com/a/47970347), which I'd be surprised if we'd want to do that.

I'm happy to start working through other tests as needed, but just want to make sure I understand the test writing conventions before going through more!

EDIT: I also accidentally created this branch off of the one in #394, so that should be merged before this.

------


gipert:
Hi Sam, I fully agree we should test processors with NumPy arrays!

Let's discuss test coverage with Numba in #241.

iguinn (commented):
Looks good! For these kinds of tests that are checking multiple different functionalities, it might be nice to have a list of the tests performed in the docstring or something like that. Having it all in one place I think will make it easier to see if there is anything missing.

Can you also test the `__wrapped__` suggestion from https://github.com/legend-exp/pygama/issues/241, and see if that triggers the coverage of this function?

gipert:
These upload failures seem to occur with tokenless uploads: https://github.com/codecov/codecov-action/issues/557#issuecomment-1216749652

It should be fixed now. The CI will fail if uploading fails, from now on.

Let's wee whether the Codecov bot shows up now.

PR #401, opened 2022-10-25
Bug fixes and improvements to the waveform browser

iguinn:
1) Fixed bug where Error wasn't being thrown when an out of range waveform index is given to the browser
2) Fixed the waveform browser's use of the default style; instead of by default drawing all waveforms blue, it will now cycle through the rcParams color cycle
3) Automatically expand wildcards, environment variables and the home directory (~) with many of our file opening functions. Added `pygama.lgdo.lgdo_utils.expand_path`, which uses glob and os.path to do its work. It also has an option for whether to ask for a single path or a list of paths; if it is a single path, it will raise an Error if there is not a single file that uniquely matches the wildcards. If it is a list of paths no Error will be raised. Reading of lh5 files (`'r'` mode only), dsp_config files, database json files, and LH5Iterators all use this now (and by extension other utilities that evoke these such as the waveform browser)
4) Fixed auto limit calculation failing with NaNs

### DIFF
--- a/src/pygama/dsp/build_dsp.py
+++ b/src/pygama/dsp/build_dsp.py
@@ -17,9 +17,9 @@ from tqdm import tqdm
 import pygama
 import pygama.lgdo as lgdo
 import pygama.lgdo.lh5_store as lh5
-from pygama.lgdo.lgdo_utils import expand_path
 from pygama.dsp.errors import DSPFatal
 from pygama.dsp.processing_chain import build_processing_chain
+from pygama.lgdo.lgdo_utils import expand_path
 
 log = logging.getLogger(__name__)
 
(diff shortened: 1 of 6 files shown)


gipert:
Hey Ian, I was wondering, can we split `lh5_store.py` in multiple files? It's pretty long now. I propose to:
- keep only `LH5Store` in `lh5_store.py`
- move `LH5Iterator` to `lh5_iterator.py`
- move functions like `ls`, `show`, `load_nda`, etc to `lh5_utils.py` (or just merge it with `lgdo_utils.py` and have only `utils.py`?)

All of them would be still imported in `__init__.py`, so little impact from the user perspective.

CC @jasondet

jasondet:
@gipert I like all of those suggestions but don't think that they need to be in this PR.

PR #402, opened 2022-10-25
Add function to run both the numba and pure python versions of some function

slwatkins:
As discussed in #241 and proposed in https://github.com/legend-exp/pygama/issues/241#issuecomment-1289774168, we can write a function that takes a numba-wrapped function, and automatically runs both the numba version and the unwrapped pure python version. For testing these numba-wrapped functions, it is necessary to run the pure python version to trigger the code coverage.

I've added a fixture to `tests/conftest.py`, which returns such a function for some generic wrapped function in, e.g., `dsp`. I've also updated the `fixed_pickoff_time` test to use this function as an example, which cleans up the test substantially.

I'll keep this as a draft for now, as I'm curious if there's a better place to store this function or if there are any thoughts on how I'm using it. I also want to point out that it doesn't work well with `pytest.raises`, so I'm just using `inspect.unwrap` for checking if the errors are being raised for both cases.

----

### DIFF
--- a/docs/source/developer.rst
+++ b/docs/source/developer.rst
@@ -5,11 +5,15 @@ The following rules and conventions have been established for the package
 development and are enforced throughout the entire code base. Merge requests
 that do not comply to the following directives will be rejected.
 
-To start developing :mod:`pygama`, clone the remote repository:
+To start developing :mod:`pygama`, fork the remote repository to your personal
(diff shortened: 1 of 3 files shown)


gipert:
Like! Can be merged from my point of view, after adding a docstring and a note to the dev's guide. I am curious to see whether vectorization will yield results which differ from the pure python implementation by more than machine's precision.

jasondet:
This is awesome @slwatkins thanks! I'll merge it once the checks are complete

PR #407, opened 2022-11-12
Added windowing processor that takes t0 and tf as parameters for data trimming

SamuelBorden:
If we choose to adopt a `build_dsp` approach to trimming raw waveforms, then we cannot use the current `windower.py` processor because that pads the array with numpy nans when a negative t0 index is used. This is a quick fix and is only intended for use in `data_trimmer.py`, which will live the legend-dataflow repo until the functionality is incorporated into pygama.

### DIFF
--- a/src/pygama/dsp/processors/__init__.py
+++ b/src/pygama/dsp/processors/__init__.py
@@ -61,6 +61,7 @@ several advantages:
 
 from .bl_subtract import bl_subtract
 from .convolutions import cusp_filter, t0_filter, zac_filter
+from .double_windower import double_windower
 from .dwt import discrete_wavelet_transform
 from .fftw import dft, inv_dft, psd
 from .fixed_time_pickoff import fixed_time_pickoff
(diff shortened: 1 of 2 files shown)


SamuelBorden:
See https://github.com/legend-exp/legend-dataflow/pull/8 for the actual processing chain that uses the `double_windower.py` processor

jasondet:
A couple of questions:

1. Why the word "double" in the name? What is double about the windower?

2. Outside of the processing framework, this can be achieved with a line like:
`waveform.values.nda = waveform.values.nda[t0:tf]`
I believe that in a line like this with numpy, no memory is reallocated; numpy just manages a "view" of the allocated memory for the slice. We should check whether this might be more performant before settling on the processing-framework-based solution.

@iguinn can you check the seaborn deprecation warning in the wf browser test?

SamuelBorden:
Closing this. It's more memory efficient to pass a view of the raw table than it is to allocate a new array with `processing_chain`. See implementation in `data_trimmer.py` here: https://github.com/legend-exp/legend-dataflow/pull/8`

PR #411, opened 2022-11-21
Add DPLMS DSP processor

valerioda:
DSP processor for LAr instrumentation SiPMs using the DPLMS method for optimum filter calculation (arXiv:2209.09863).
This processor is composed of a factory function "dplms_filter" that is called using the `init_args` argument. The input and output waveforms are passed using `args`.
The function "dplms_filter" input arguments are:
- noise_mat: list noise matrix (from the database)
- reference: list reference signal (from the database)
- length: int filter length
- a1: float penalized coefficient for the noise matrix
- a2: float penalized coefficient for the reference matrix
- a3: float penalized coefficient for the zero area matrix
- ff: intflat top length for the reference signal
A test config file for SiPM processing has been included.


jasondet:
Add a test please.

gipert:
> Add a test please.

See https://pygama.readthedocs.io/en/latest/developer.html#testing-numba-wrapped-functions

valerioda:
After PR #425 possible to add a test for dplms numba processor, running successfully
`pytest tests/dsp/processors/test_dplms.py`

gipert:
Looks great! I'll merge once GitHub checks are through

PR #414, opened 2022-11-21
Added unnormalized presum processor that stores to ints

SamuelBorden:


SamuelBorden:
I wasn't sure if we wanted to store as signed or unsigned int, so I left the option for both, and the user can decide in their dsp config

gipert (changes requested):
Hi Sam, I would propose to merge this into the `presum` processor. There could be for example a flag to specify whether there should be a normalization (maybe we need to turn `presum` in a factory then)

SamuelBorden:
Closing this merge, updating the `presum` processor in #415

PR #422, opened 2022-12-16
Switch to GPLv3 license

gipert:
I had a look at which license we could use for pygama (and, in general, for all our open source software). Among the most common licenses, [GPL3](https://choosealicense.com/licenses/gpl-3.0) looks like the one we need, in my opinion. It is a bit more restrictive compared to the Apache one, in terms of conditions (check out this nice [summary](https://choosealicense.com/appendix/)).

If people have a different opinion and want to be more permissive, we could use [BSD3](https://choosealicense.com/licenses/bsd-3-clause/).


jasondet:
ok by me...

PR #425, opened 2022-12-20
Add functionality for testing factory functions in DSP, Add discrete_wavelet_transform test

slwatkins:
I've updated the fixture `compare_numba_vs_python` to work with factory functions for making processors in `dsp`. Since the function signature is a special case for these, the current version of the fixture was not working.

In this PR, I've added:
 - A check of the function signature in `compare_numba_vs_python` for if a value is returned by the function (as opposed to updated in-place)
   - If True, the previously written version is run (no changes here)
   - If False, then a pre-initialized output must be passed as an input, which will then be changed in-place and the Numba and Python versions of the function will be properly compared
- A test for `discrete_wavelet_transform` as a simple example for testing one of the factory functions in the processors
- Updated the developer documentation to mention this way of testing these cases

This was done to help #411 write the needed tests, as this new processor relies on a factory function. With this change, it should be easy to write the test for that processor.

-----

### DIFF
--- a/docs/source/developer.rst
+++ b/docs/source/developer.rst
@@ -192,6 +192,59 @@ required imports in the snippet for verbosity.
 In this case, the general idea is to use :func:`pytest.raises` twice, once with
 ``func(*inputs)``, and again with ``inspect.unwrap(func)(*inputs)``.
 
+Testing Factory Functions that Return Numba-Wrapped Functions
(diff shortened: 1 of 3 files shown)


gipert:
Looks great! Thanks a lot, I'll merge this right away.

PR #426, opened 2022-12-22
Fix check for if the decoded values were built in ORFlashCamADCWaveformDecoder

slwatkins:
I noticed that the check for if the `decoded_values` were built when trying to access them in `ORFlashCamADCWaveformDecoder` had a bug:

- `dict_values` is not subscriptable
- It would try to index into the list even if the list length was zero

So, I made a small fix to have the behavior match what was expected.

Here's a MWE:

```python
from pygama.raw.orca.orca_flashcam import ORFlashCamADCWaveformDecoder

ORFlashCamADCWaveformDecoder().get_decoded_values()
```

-----


gipert:
Thanks!

PR #421, opened 2022-12-07
Add ORRunDecoderForRun to Orca decoders

slwatkins:
I wrote up an initial version of the ORRunDecoderForRun, where the data format was taken from http://orca.physics.unc.edu/orca/Data_Chain/Run_Control.html. I've tested it with build_raw on the L200 test data, and it appears to fit well into the workflow (no errors, and the data shows as expected in the outputted LH5 file).

I started this as a draft PR, since I'm not totally clear how exactly we want to write up new decoders and have them fit into pygama.

A couple of quick comments on basic things I was thinking about when writing up this simple decoder:

- Did I put it in the right spot? (i.e. should it live in orca_flashcam.py, or somewhere else?)
- Is the structure okay? I just based the structure on the ORFlashCamListenerConfigDecoder, since they seemed to both be relatively simple decoders
- I arbitrarily named various decoded values, are there specific conventions that I should follow?

I need to add docstrings, and I'm guessing I should add a test as well (or is there a test I should update?).

---------
Before submitting a pull request, please make sure you've read and understood the pygama developer's guide: https://pygama.readthedocs.io/en/latest/developer.html. In particular, do not forget to:

- [x] Conform to our **coding conventions**
- [x] Update existing or add new **tests**
- [x] Update existing or add new **documentation**
- [x] Address any issue reported by GitHub checks

PR #427, opened 2022-12-22
Adding daq to raw validation code

slwatkins:
I've written up a first draft of an option for the daq to raw validation code. Right now it's written all in one test file, which makes an LH5 file via `build_raw` on a LEGEND test data file, recreates the orca file from the LH5 file, and then does a direct comparison to the original orca file.

I was hoping to get some feedback on this, I think a few questions are:
 - Should the `OrcaEncoder` live elsewhere?
   - Someone may want to use this to create "fake" orca files
 - Should we make a new test data file with data IDs other than those currently being supported by pygama?
   - If so, we'd need to add the orphaned packets functionality to `build_raw` (likely relatively simple)
   - Or we would need to write up the rest of the decoders (probably would take a while)
 - Any other thoughts on the validation?

I'm going to leave this PR as a draft, as there's likely a fair amount of work left - mostly wanted to start a conversation about the current state of the validation. FYI, this branch was made off of the one related to #421.

-----

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -7,6 +7,7 @@ on:
       - refactor
       - 'releases/**'
   pull_request:
+  merge_group:
   release:
 
 concurrency:
--- a/src/pygama/lgdo/lh5_store.py
+++ b/src/pygama/lgdo/lh5_store.py
@@ -307,7 +307,7 @@ class LH5Store:
         if datatype == "scalar":
             value = h5f[name][()]
             if elements == "bool":
-                value = np.bool(value)
+                value = np.bool_(value)
             if obj_buf is not None:
                 obj_buf.value = value
                 obj_buf.attrs.update(h5f[name].attrs)
@@ -631,7 +631,7 @@ class LH5Store:
             # special handling for bools
             # (c and Julia store as uint8 so cast to bool)
             if elements == "bool":
-                nda = nda.astype(np.bool)
+                nda = nda.astype(np.bool_)
 
             # Finally, set attributes and return objects
             attrs = h5f[name].attrs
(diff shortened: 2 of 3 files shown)


slwatkins:
Looks like the tests failed because we use `np.bool` instead of `np.bool_` in `lh5_store.py`, which gives a deprecation warning 🤷 

https://github.com/legend-exp/pygama/blob/f5220da351419cfc8276e794fb3c9595cbc6fb1d/src/pygama/lgdo/lh5_store.py#L310

https://github.com/legend-exp/pygama/blob/f5220da351419cfc8276e794fb3c9595cbc6fb1d/src/pygama/lgdo/lh5_store.py#L634

jasondet:
@slwatkins is this still a draft or is it ready for review?

slwatkins:
I can mark it as ready for review, but I think it needs some changes/input on where various pieces of code goes and what is missing in this PR.

gipert:
So the idea of this test would be to make sure that no information from the original FlashCam file is lost in translation, right? It cannot be a reliable test of whether the raw data is also encoded in the right way, I guess, since one would have to validate `OrcaEncoder` too.

slwatkins:
My understanding was that the purpose of this test is to check if we are losing any information when converting an ORCA file to a LH5 file, and the way to do this is to reconstruct the ORCA file bit-by-bit from an LH5 and do a direct comparison. So, in that way I feel that this validates both the decoders and the "encoder" (which i wrote just as a helper function for this, but I could see a use for creating "fake" ORCA files, which is why we may want to generalize it).

Validating FlashCam <-> ORCA I believe falls under the purview of the DAQ group, correct?

gipert:
> Validating FlashCam <-> ORCA I believe falls under the purview of the DAQ group, correct?

Of course, I was referring to unit tests for the `raw.orca` subpackage which are still missing and we should add at some point (like the ones we already have for `raw.fc`). But checking that no information is leaking is already a very important test!

This PR can be merged from my point of view.

PR #428, opened 2022-12-29
LH5 code reorganization

### LINKED ISSUE #406: `lh5_store.py` refactor
We should split `lh5_store.py` in multiple files, It's pretty long now. I propose to:

- keep only `LH5Store` in `lh5_store.py`
- move `LH5Iterator` to `lh5_iterator.py`
- move functions like `ls`, `show`, `load_nda`, etc to `lh5_utils.py` (or just merge it with `lgdo_utils.py` and have only `utils.py`?)

All of them would be still imported in `__init__.py`, so little impact from the user perspective.

gipert:
The LH5 codebase has expanded significantly in the last year or so: it's time for a refactor. Since LH5 is a codification of LGDOs into HDF5 format, I propose to move it in a dedicated subpackage of `lgdo` called `lh5`. This implies the following import statement syntax:
```py
from pygama.lgdo import Table, Array # as before
from pygama.lgdo.lh5 import LH5Store, ls # new
```

The question is whether we also want to rename (`LH5Store`, `LH5Iterator`) -> (`Store`, `Iterator`) at this point, since they live in the `lh5` subpackage.

Summary of changes:

* [x] Moved LH5 code into own subpackage `lgdo.lh5` and renamed modules. This is how it looks now:
   ```
   src/pygama/lgdo
   ├── arrayofequalsizedarrays.py
   ├── array.py
   ├── fixedsizearray.py
   ├── __init__.py
   ├── lh5
   │   ├── __init__.py
   │   ├── iterator.py
   │   ├── store.py
   │   └── utils.py
   ├── scalar.py
   ├── struct.py
   ├── table.py
   ├── utils.py
   ├── vectorofvectors.py
   └── waveform_table.py
   ```
* [x] Created `lgdo.utils` and `lgdo.lh5.utils`
* [x] Update all import statements
* [ ] Update documentation (maybe add `-n` flag to `sphinx-build` to check cross references)

**Note:** this PR breaks backward compatibility but only affects package structure and naming. Users will in most cases only need to update their import statements.

Closes #406.


gipert:
> HDF5 should be seen as one option for the on-disk persistent format of LGDO's. Even if the inspiration was the other way around, and no other option is currently available.

Indeed. We could for example have at some point a `g4h5` submodule that reads Geant4 HDF5 files into LGDOs.

We could also have all compression-related stuff inside a `compression` submodule.

I will rebase this PR once #442 is closed.

gipert:
Will rebase after #442 is merged. Closing for now.

PR #433, opened 2023-01-09
`processors.get_multi_local_extrema()`: new advanced search methods (+ unit tests)

patgo25:
Added Features to get_multi_local_extrema processor:

- Choose the search direction:
    -  Left to right
    -  Right to Left
    -  Both directions conservative (find the smallest mutual set of extrema)
    -  Both directions aggressive (find the largest set of unique extrema)
- Set different delta limits for minima and maxima search
- Updated tests

### DIFF
--- a/src/pygama/dsp/processors/get_multi_local_extrema.py
+++ b/src/pygama/dsp/processors/get_multi_local_extrema.py
@@ -170,8 +170,6 @@ def get_multi_local_extrema(
                     imax = i
                     find_max = True
 
-    else:
-        raise DSPFatal("search direction type not found.")
     # set output
     # left search
     if search_direction == 0:
--- a/src/pygama/dsp/processors/get_multi_local_extrema.py
+++ b/src/pygama/dsp/processors/get_multi_local_extrema.py
@@ -136,7 +136,7 @@ def get_multi_local_extrema(
                     find_max = True
 
     # right to left search
-    elif search_direction > 0:
+    if search_direction > 0:
         find_max = True
         imax, imin = len(w_in) - 1, len(w_in) - 1
         for i in range(len(w_in) - 1, -1, -1):
--- a/src/pygama/dsp/processors/get_multi_local_extrema.py
+++ b/src/pygama/dsp/processors/get_multi_local_extrema.py
@@ -6,8 +6,6 @@ from numba import guvectorize
 from pygama.dsp.errors import DSPFatal
 from pygama.dsp.utils import numba_defaults_kwargs as nb_kwargs
 
-# from numba.types import bool_, int_
-
 
 @guvectorize(
     [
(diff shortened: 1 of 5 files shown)


gipert:
Thanks @patgo25! Can you also add a couple of simple unit tests in tests/dsp/processors/test_get_multi_local_extrema.py to this PR?

jasondet:
A couple of questions:
- can you describe in what sort of situations L->R might give a different result from R->L?
- can you describe in what sort of situations one would want different deltas for minima / maxima?
- have you verified that L->R with one delta reproduces the original behavior?

patgo25:
Added some tests @gipert 

@jasondet, the answer to your question is also tested in the tests. But here are the fast answers:
- for symmetric deltas, the search direction doesn't matter
- for asymmetric deltas, it matters:
```
         2
     1   /\  3
     /\/    \/\
    /  a     b \
___/            \____
```
                                             
delta_min =delta_max  is (a little less than) two lines high.
L -> R = L <- R = L<-c-> R = L<-a->R:  Recognized maxima: [2], recognized minima []

delta_min is (a little less than) one line high, and delta_max  is (a little less than) two lines high.
L -> R:  Recognized maxima: [2,3] , recognized minima [b]
L <- R:  Recognized maxima: [2,1] , recognized minima [a] 
L<-c-> R: Recognized maxima: [2] , recognized minima []
L<-a-> R: Recognized maxima: [1,2,3] , recognized minima [a,b]

sorry for my bad ASCII art. ;)

- Specifically, I write this for the SiPM analysis. But this is a little too hard to express in ASCII art ;) we can talk about it somewhere where I can send you actual plots :)
- Generally, you could achieve the thing in the ASCII "plot" above with a smaller symmetric delta. But let's say you have something like this:
```
          2
          /\      4
     1  /   \ 3  /\
     /\/     \/\/  \
    /  a     b  c   \
___/                 \____
```
With a small symm. delta (one line high), you would identify all minima and all maxima.
With an asym delta d_max = 2 lines, d_min = 1 line height, you would get:
L -> R:  Recognized maxima: [2,4] , recognized minima [b]
L <- R:  Recognized maxima: [2,1] , recognized minima [a] 
L<-c-> R: Recognized maxima: [2] , recognized minima []
L<-a-> R: Recognized maxima: [1,2,4] , recognized minima [a,b]
This way, one would mitigate the extrema detection of [c,3]

- L->R with symm. delta reproduces original behavior.

jasondet:
Very nice! And the ascii art was very impressive: effective and helpful! 

I'm all for this. It would be good to have someone more expert like @iguinn or @SamuelBorden or @ggmarshall take a quick look. If any of them gives the thumbs up I'll merge.

gipert:
I reviewed the changes with @patgo25 and they look OK to me. I'll merge if tests pass.

gipert:
We keep getting the Codecov error, that's a known issue but we cannot do anything about that.

PR #434, opened 2023-01-09
`processors.histogram()`: check that bin edges array is +1 longer than weights array

### LINKED ISSUE #388: Histogram processor: No check of array sizes
The histogram processor should check if the array size of borders is precisely one larger than the bin array size.

Everything else makes no sense anyhow and leads to strange crashes.

patgo25:
fix #388


gipert:
Thanks @patgo25! Can you also add a couple of simple unit tests in `tests/dsp/processors/test_histogram.py` to this PR?

gipert:
(Ignore the failing CI, it's not our fault)

PR #443, opened 2023-01-19
Created a temporary table for buffer_processor to write to file during build_raw

SamuelBorden:
This update fixes the unwanted behavior of the `buffer_processor` dropping fields in a buffer before `build_raw` is done writing to the buffer. The major change is that `buffer_processor.py` now returns a temporary `lgdo.Table` that shares columns with an `rb.lgdo`, and all processing is done on this temporary table instead of directly on the `rb.lgdo`.


jasondet:
looks perfect.

PR #444, opened 2023-01-20
Added min_max normalization

esleon97:


gipert:
Ciao @esleon97, could you describe a use case?

esleon97:
Hi @gipert. I use this to preprocess waveforms before feeding them into my machine learning data cleaning model. This processor ensures that the inputs are constrained to the [-1,1] range, which makes the model perform better.

esleon97:
Hi @gipert. The data cleaning model mostly relies on other Python machine learning packages. I am using pygama to preprocess waveforms (perform discrete wavelet transforms and max/min normalization). I added this processor to have the preprocessing stage done in a single framework while taking advantage of pygama's dsp vectorization. However, I could also incorporate this function without using pygama.

gipert:
> I added this processor to have the preprocessing stage done in a single framework while taking advantage of pygama's dsp vectorization.

The vectorization is done by Numba, so having your processor in a separate package does not introduce any performance penalty -- as long as you decorate it with `@guvectorize`.

For the time being, I would recommend to keep it in your package, let's discuss this again once the code is ready to be integrated ;)

PR #448, opened 2023-01-26
DataLoader: handling of non-scalar data

gracesong312:
Main updates to functionality:
1. WaveformTable (and general Table) loading
- Used to: manually explode columns of a WaveformTable so that one WaveformTable got loaded as three Arrays (for `dt`, `t0`, and `values`)
- Now: Load Tables as Tables, only explodes columns if the user requests a DataFrame, which it does in `get_dataframe` 
2. ArrayOfEqualSizedArray and VectorOfVector loading
AOESA will get loaded as VoV because the size of an array across multiple channels may be different. Added a constructor to VoV that builds a VoV out of a list of lists in order to accommodate this. 
3. Support `sortby` in FileDB config
Currently, FileDB sorts the dataframe by the key "timestamp," which is fine for L200, but test stands may not use a naming convention that includes "timestamp", so the user can now specify which key should be used for sorting. It still defaults to "timestamp" if no `sortby` is specified. 

For testing, I'm not entirely sure what's failing in the codecov and docs, I really don't think I changed anything. Also, I've tested VoV loading separately, but there isn't really good VoV data in the test suite, so I haven't added that in the tests folder.


src/pygama/flow/file_db.py
```diff
@@ -192,6 +192,11 @@ def set_config(self, config: dict, config_path: str = None) -> None:
         self.tier_dirs = self.config["tier_dirs"]
         self.table_format = self.config["table_format"]
 
+        if "sortby" in self.config.keys():
+            self.sortby = self.config["sortby"]
+        else:
+            self.sortby = "timestamp"
+
```
gipert, line 199: This can be one line using `dict.get(key, default_value)`

src/pygama/lgdo/vectorofvectors.py
```diff
@@ -55,35 +59,48 @@ def __init__(
...
         """
-        if cumulative_length is None:
-            self.cumulative_length = Array(
-                shape=(shape_guess[0],), dtype="uint32", fill_val=0
+        if listoflists is not None:
+            cum_length = []
+            for vec in listoflists:
+                cum_length.append(len(vec))
+            self.flattened_data = Array(
+                nda=np.array([item for sublist in listoflists for item in sublist])
```
gipert, line 71: Python for loops with built-in lists... Can this be made more efficient with Numpy?

gipert (changes requested):
I noticed some functions without type hints, could you add them please?

src/pygama/lgdo/vectorofvectors.py
gracesong312: So the issue is that numpy doesn't like jagged arrays. I'm not sure what an efficient way around that would be.

jasondet: I think I found a more efficient way. It should be good enough for us.

gipert: In the long term we'll want to have more interoperability between `VectorOfVectors` and [`awkward.Array`](https://awkward-array.org/doc/main/index.html). Once #377 is addressed, we could support passing an awkward array as `listoflists`, that should be the fastest.

gipert:
pre-commit.ci run

gipert:
The pre-commit.ci failure seems an upstream problem to me, let's see if they fix it by today.

gipert (approved):

jasondet:
@gipert can you check the error?

jasondet:
I figured it out.

PR #449, opened 2023-01-27
Fix for scipy bug/change

ggmarshall:
Erin found a bug yesterday at nersc where the scipy minimize function was returning a float instead of an array. I managed to recreate a minimal version of this:

```python
In [1]: import numpy as np

In [2]: from scipy.optimize import minimize

In [3]: import scipy

In [4]: scipy.__version__
Out[4]: '1.7.1'

In [5]: def test_func(x):
   ...:    return np.array([x**2]).reshape(1,)
   ...: 

In [6]: response = minimize(fun=test_func, x0=np.array([0.5]).reshape(1,), bounds=[(-1, 1)],method="L-BFGS-B")

In [7]: response.fun
Out[7]: array([7.70371947e-18])

In [1]: import numpy as np

In [2]: from scipy.optimize import minimize

In [3]: import scipy

In [4]: scipy.__version__
Out[4]: '1.10.0'

In [5]: def test_func(x):
   ...:     return np.array([x**2]).reshape(1,)
   ...: 

In [6]: response = minimize(fun=test_func, x0=np.array([0.5]).reshape(1,), bounds=[(-1, 1)],method="L-BFGS-B")

In [7]: response.fun
Out[7]: 7.703719366273389e-18
```

This seems to have been caused by a change in scipy version 1.8 where arrays of shape (1,) are now cast to floats. To fix this I have changed the BayesianOptimiser's acquisition functions so that they will now output a float in all cases. I have also added a minimum version of scipy at 1.0.1 as this is when  OptimiseResult was added. I'm not sure if any other parts of pygama need a newer version.

### DIFF
--- a/setup.cfg
+++ b/setup.cfg
@@ -45,7 +45,7 @@ install_requires =
     pyfcutils
     pyfftw
     scikit-learn
-    scipy
+    scipy>=1.8.1
     tables
     tqdm
 python_requires = >=3.9
--- a/setup.cfg
+++ b/setup.cfg
@@ -45,7 +45,7 @@ install_requires =
     pyfcutils
     pyfftw
     scikit-learn
-    scipy>=1.8.1
+    scipy>=1.0.1
     tables
     tqdm
 python_requires = >=3.9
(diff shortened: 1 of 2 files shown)


gipert:
> I'm not sure if any other parts of pygama need a newer version.

still better than nothing!

Issue #451, opened 2023-02-02
Implement is_group check to build_dsp


MoritzNeuberger:
There seems to be a bug with gimme_file().
When lh5_file name is not the name of a group in the file, gimme_file() returns a dataset instead of a file/group.
To fix this, we have to add a "is group" check to build_dsp(): 
https://github.com/legend-exp/pygama/blob/main/src/pygama/dsp/build_dsp.py#L110

MoritzNeuberger:
Hmm, I thought I could automatically generate a branch from the issue but seem to be unable to do that.

PR #452, opened 2023-02-02
first attempt to a is not dataset check (addressing issue #451)

MoritzNeuberger:
Before looking for a nested group in `"{tb}/raw"`, we check whether the table entry `"{tb}"` in the file is a dataset or not.
If it is, the table entry is deleted because it cannot contain a raw group.

### DIFF
--- a/src/pygama/dsp/build_dsp.py
+++ b/src/pygama/dsp/build_dsp.py
@@ -123,7 +123,7 @@ def build_dsp(
         if (
             "raw" not in tb
             and not isinstance(
-                raw_store.gimme_file(lh5_file, "r")[f"{tb}"], h5py.Dataset
+                raw_store.gimme_file(lh5_file, "r")[tb], h5py.Dataset
             )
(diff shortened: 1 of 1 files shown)


src/pygama/dsp/build_dsp.py
```diff
@@ -120,7 +120,13 @@ def build_dsp(
 
     # check if group points to raw data; sometimes 'raw' is nested, e.g g024/raw
     for i, tb in enumerate(lh5_tables):
-        if "raw" not in tb and lh5.ls(lh5_file, f"{tb}/raw"):
+        if (
+            "raw" not in tb
+            and not isinstance(
+                raw_store.gimme_file(lh5_file, "r")[f"{tb}"], h5py.Dataset
```
gipert, line 126:
```suggestion
                raw_store.gimme_file(lh5_file, "r")[tb], h5py.Dataset
```

gipert:
Thank you Moritz

Issue #176, opened 2021-07-30
Multiple store.write_object calls


sweigart:
When the `write_object` function of the `Store` class is run multiple times, I find that the "datatype" attribute of the HDF5 group is not updated after the first function call.  This causes a subsequent `read_object` call to read out only what was written with the first `write_object` call.  Here's an example:

```
>>> import pygama.lh5 as lh5
>>> import h5py
>>> import numpy as np
>>> tb1 = lh5.Table(col_dict={'dset1': lh5.Array(np.zeros(10))})
>>> tb2 = lh5.Table(col_dict={'dset2': lh5.Array(np.zeros(10))})
>>> store = lh5.Store()
>>> store.write_object(tb1, 'my_group', 'my_output.lh5')
>>> store.write_object(tb2, 'my_group', 'my_output.lh5')
>>> tb_dat, _ = store.read_object('my_group', 'my_output.lh5')
>>> tb_dat.keys()
dict_keys(['dset1'])
>>> f = h5py.File('my_output.lh5', 'r')
>>> f.keys()
<KeysViewHDF5 ['my_group']>
>>> f['my_group'].attrs['datatype']
'table{dset1}'
```

Here only the `dset1` dataset was read with `read_object` call, and, by opening the file directly, we can see that this is because only `dset1` is in "datatype" attribute for the HDF5 group.  I would have expected that `dset2` also be in the "datatype" attribute (and that `dset2` be also read in with a `read_object` call).  I would make a pull request to create such behavior in store.py, but I do not currently understand enough of the class' code to do so.

Now, in the above example, I would get the desired result if I add `dset2` to the `col_dict` argument in `lh5.Table` when creating the `tb1` object.  However, my use case requires me to add new datasets to the same group in separate steps, so such a fix is not possible.  I can also imagine a use case where new datasets are added to the same group using `write_object` using different scripts, which runs into this same issue when `read_object` is run with the name of the group.

jasondet:
Jason take a look in refactor

SamuelBorden:
I've tried taking a look at this, especially after the `lgdo` refactor, to see if this is still an issue. Running the same block of code as above results in the following: 
``` 
>>> import pygama.lgdo as lh5 
>>> import h5py
>>> import numpy as np 
>>> tb1 = lh5.Table(col_dict={'dset1': lh5.Array(np.zeros(10))})
>>> tb2 = lh5.Table(col_dict={'dset2': lh5.Array(np.zeros(10))})
[...]

PR #455, opened 2023-02-10
Enhancement: ability to append columns to existing tables in lh5 files using `store.write_object`

SamuelBorden:
This addresses the enhancement part of issue #176. Users can now append new columns to an existing `lgdo.Table` in an `lh5` file by using the `append_column` argument of the `wo_mode` parameter. 

If the user tries to `append_column` with a table that has any of the same fields as the existing table, or if the new table is a different size than the existing table, then the function will raise an error.  

```
>>> import pygama.lgdo as lh5 
>>> import numpy as np 
>>> tb1 = lh5.Table(col_dict={'dset1': lh5.Array(np.zeros(10))})
>>> tb2 = lh5.Table(col_dict={'dset2': lh5.Array(np.ones(10))})  # second table with new field name
>>> store = lh5.LH5Store()
>>> store.write_object(tb1, name='my_table', lh5_file="my_output.lh5")
>>> store.write_object(tb2, name='my_table', lh5_file="my_output.lh5", wo_mode="append_column")
>>> tb_dat, _ = store.read_object("my_table", "/tmp/my_output.lh5")
>>> tb_dat['dset1'].nda
[0. 0. 0. 0. 0. 0. 0. 0. 0. 0.]
>>> tb_dat['dset2'].nda
[1. 1. 1. 1. 1. 1. 1. 1. 1. 1.]
```

### DIFF
--- a/src/pygama/lgdo/lh5_store.py
+++ b/src/pygama/lgdo/lh5_store.py
@@ -115,7 +115,7 @@ class LH5Store:
         grp_attrs
             HDF5 group attributes.
         overwrite
-            whether overwrite group attributes, ignored is `grp_attrs` is
+            whether overwrite group attributes, ignored if `grp_attrs` is
             ``None``.
         """
         if not isinstance(group, h5py.Group):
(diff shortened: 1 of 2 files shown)


gipert:
Great! Thanks a lot @SamuelBorden.

PR #456, opened 2023-02-15
update tcm documentation

jasondet:


src/pygama/evt/tcm.py
```diff
@@ -42,11 +44,9 @@ def generate_tcm_cols(
     window_ref
         when testing one datum for inclusion in a cluster, test if it is within
         `coin_window` of
-
```
gipert, line 45: without newlines before and after lists these will not render properly on the web docs

jasondet: the Build Documentation check made the same complaint

PR #458, opened 2023-03-07
Fix hardcoded dtype for output buffer

erin717:
The dtype for the UnitConversionManager out_buffer was hardcoded to be a float64, causing bugs with some processors. This is fixed by changing the dtype to be var.dtype.

### DIFF
--- a/src/pygama/dsp/processing_chain.py
+++ b/src/pygama/dsp/processing_chain.py
@@ -1250,7 +1250,7 @@ class UnitConversionManager(ProcessorManager):
     def convert_int(buf_in, offset_in, offset_out, period_ratio):  # noqa: N805
         tmp = (buf_in + offset_in) * period_ratio - offset_out
         ret = round(tmp)
-        if np.abs(tmp-ret)<1.e-5:
+        if np.abs(tmp - ret) < 1.0e-5:
             return ret
         else:
             raise DSPFatal(f"Cannot convert to integer")
--- a/src/pygama/dsp/processing_chain.py
+++ b/src/pygama/dsp/processing_chain.py
@@ -1257,7 +1257,7 @@ class UnitConversionManager(ProcessorManager):
 
         from_buffer, from_grid = var._buffer[0]
         period_ratio = from_grid.get_period(unit.period)
-        self.out_buffer = np.zeros_like(from_buffer, dtype="float64")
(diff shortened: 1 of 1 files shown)


gipert:
Hi @erin717, thanks a lot for the fix! The DSP tutorial notebook somehow broke -- could you investigate?

ggmarshall:
Can I ask which processors were having issues? I think we decided a while ago that we wanted the output of all processors to be floats so that if there were issues the output would just be nan

iguinn:
The int32 thing is happening for `np.argmax`, which returns an integer but is sometimes used for times. For pygama processors, we absolutely should continue to use floats for the reason you said, but we also need to handle the case where an integer is used. However, the initial problem was caused by trying to use float32 for times (which we want to be allowed)

To give a bit more detail, what's happening is that when we convert from one coordinate system to another (which happens when we do windowing or down-sampling), time values have to be automatically adjusted if they are measured in one coordinate system and then passed to another. The processor for doing this was just converting everything to `float64`. This was causing problems for the windower, because this was causing it to think it was getting a `float32` when it was really getting a `float64`, which caused it to want to automatically cast `wf_out` from `float32` to `float64`. This is a problem in the very special case where the output has a different shape than the input, because in this case, when we set up the signature for the ufunc, it treats `wf_out` as an input, and only does this cast when it is passed in, and doesn't cast back when it goes out.

The real problem here was that the processing chain thought it was passing a `float32` when it was actually passing a `float64`. The solution to this is to make sure that the unit converter keeps time values as the same dtype as they started with, which means _not_ hardcoding it as `float64`. That's the initial fix that Erin put in.

This, however, runs into a problem, specifically with `int`s (like `argmax`), where convert raises an error because the operations inside of it want to return a `float` instead of an `int`. This error is the reason I hardcoded it as `float64` in this pull request https://github.com/legend-exp/pygama/pull/364. Now we have to bug fix that bug fix.

The solution, which Erin is currently testing, is to have a special case for `int`s where it rounds and returns an `int` if the converted timing value is close to an integer, and raises an error if the timing value is not close to an integer. If someone runs into this error (which could happen with downsampling and argmax being used together), the solution would be to explicitly have argmax return into a `float`.

TL;DR: Converting timing values between different coordinate systems had two poorly handled edge-cases, which Erin is fixing

PR #459, opened 2023-03-12
New Orca/FlashCam keying convention, fix timestamp issue, draft status decoding

jasondet:


jasondet:
Looks like the tests are finally passing. @mmatteo and @ggmarshall I'm certain you'll need to make some changes in the processing scripts for this. At the very least the table names all need 7 digits instead of 3.

gipert:
Hi Jason, we should add the new field to the channel map. Can you recommend a short and descriptive name?

jasondet:
@gipert how about "FBI"?

gipert:
What if we use a name that makes it clear that this is the identifier used for channels in raw files? So instead of stating "what it is", we would state "how it is used". What do you think?

gipert:
For example, `rawid`.

PR #460, opened 2023-03-13
Remove deprecated `flow.DataGroup`

gipert:
The tool is deprecated and people still using it should switch to `flow.DataLoader`. Otherwise, they should maintain a pygama fork with `DataGroup` still in.


gipert:
If no objections, I will merge this PR soon.

PR #463, opened 2023-03-15
Changes ready for p03

ggmarshall:
dsp - Changed all convolution processes from np.convolve to scipy.signal.fftconvolve gave >4x speedup in testing

A/E - Can now fail gracefully at more stages 

Cuts - Rewrote how the cuts are found to be much more robust than before, also tuned pulser finding based on p02 data

Energy Optimisation - Added fallback to Minos fitting, improved event selection by including SEP, added errors to bayesian optimiser if nan the value will still contribute to gpr fit but will not be returned as optimal

Ecal - Moved into class and redefined the plotting interface can now pass a dictionary with the function to plot and any options you want to pass. Better defined failure modes such as when FEP fails to fit.

Extract_tau - improved output plots.

In all cases the pulser was changed to remove all events in window to make sure there is no contamination. I have also done a sweep through of pargen logging and lowered unnecessary warnings to info.

### DIFF
--- a/src/pygama/dsp/processors/convolutions.py
+++ b/src/pygama/dsp/processors/convolutions.py
@@ -4,6 +4,7 @@ from typing import Callable
 
 import numpy as np
 from numba import guvectorize
+from scipy.signal import fftconvolve
 
 from pygama.dsp.errors import DSPFatal
 from pygama.dsp.utils import numba_defaults_kwargs as nb_kwargs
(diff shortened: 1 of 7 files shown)


gipert:
> dsp - Changed all convolution processes from np.convolve to scipy.signal.fftconvolve gave >4x speedup in testing

Interesting! Maybe @iguinn wants to have a look?

PR #472, opened 2023-03-28
DataLoader - attribute and error handling

gracesong312:
This fixes a previous issue where attributes such as units weren't being loaded along with the data
In doing this, found a bug in how AOESA handles default dimensions and how DL was previously handling those dimensions. 

It also essentially allows the FileDB to be wrong i.e. if it expects a column or table it does not actually find in the file, it emits a warning and moves on rather than failing. 

I'm not sure why I keep getting codecov failures or how to fix them

### DIFF
--- a/src/pygama/flow/data_loader.py
+++ b/src/pygama/flow/data_loader.py
@@ -1012,7 +1012,7 @@ class DataLoader:
                     nda = np.array(col_dict[col])
                     if len(nda.shape) == 2:
                         dt = attr_dict[col]["datatype"]
-                        g = re.match("\w+<(\d+),(\d+)>{\w+}", dt).groups()
+                        g = re.match(r"\w+<(\d+),(\d+)>{\w+}", dt).groups()
                         dims = [int(e) for e in g]
                         col_dict[col] = ArrayOfEqualSizedArrays(
                             dims=dims, nda=nda, attrs=attr_dict[col]


src/pygama/flow/data_loader.py
```diff
@@ -940,35 +980,48 @@ def fill_col_dict(
...
+                    col_dict[col] = dict_to_table(
+                        col_dict=col_dict[col], attr_dict=attr_dict[col]
+                    )
                 else:
                     nda = np.array(col_dict[col])
                     if len(nda.shape) == 2:
-                        col_dict[col] = ArrayOfEqualSizedArrays(nda=nda)
+                        dt = attr_dict[col]["datatype"]
+                        dims = dt[dt.index("<") + 1 : dt.index(">")]
+                        dims = [int(e) for e in dims.split(",")]
```
gipert, line 1012: Also nice to use a regex here, something like `re.match("\w+<(\d+),(\d+)>{\w+}", dt)` (untested)

gipert:
> I'm not sure why I keep getting codecov failures or how to fix them

It randomly fails, but it's not our fault. Don't worry, doesn't blocking merging anyways

src/pygama/flow/data_loader.py
gipert: Ah sorry, I should have mentioned that the advantage is that by calling `.groups()` on the matched object you can get directly a tuple `('m', 'n')` (your `dims`)

gipert:
```suggestion
                        g = re.match("\w+<(\d+),(\d+)>{\w+}", dt).groups()
                        dims = [int(e) for e in g]
```

src/pygama/flow/data_loader.py
```diff
@@ -940,35 +980,54 @@ def fill_col_dict(
...
                 else:
+                    # ndas are Arrays or AOESA
                     nda = np.array(col_dict[col])
                     if len(nda.shape) == 2:
-                        col_dict[col] = ArrayOfEqualSizedArrays(nda=nda)
+                        dt = attr_dict[col]["datatype"]
+                        dims = re.search(r"<(\d+),(\d+)>", dt)[0][
+                            1:-1
+                        ]  # end up with 'n,m'
+                        dims = [int(e) for e in dims.split(",")]
```
gipert, line 1018:
```suggestion
                        g = re.match("\w+<(\d+),(\d+)>{\w+}", dt).groups()
                        dims = [int(e) for e in g]
```
sorry, I added the suggestion to an old commit

src/pygama/flow/file_db.py
```diff
@@ -200,7 +200,7 @@ def set_config(self, config: dict, config_path: str = None) -> None:
         # Relative paths are interpreted relative to the configuration file
         if not data_dir.startswith("/"):
             config_dir = os.path.dirname(config_path)
-            data_dir = os.path.join(config_dir, data_dir)
+            data_dir = os.path.join(config_dir, data_dir.lstrip("/"))
```
jasondet, line 203: this may fail if a user doesn't end an entry in the fdb_config.json with a '/' -- better to leave it off and let the user remove it in the json

jasondet:
looks good to me. I'm going to pull in the recent changes in main in preparation to merge.

src/pygama/flow/data_loader.py
jasondet: @gipert is this change necessary?

gipert: nope, please go ahead

Issue #492, opened 2023-05-15
DataLoader feature requests


iguinn:
Pygama now has a DataLoader that has been extremely helpful in accessing data across multiple runs and different tiers. As this tool has gotten more users, people have also had ideas for new features that will be useful in the future for a variety of analysis tasks. This issue will be used to compile a list of ideas; as people have new ideas, post them here, and I will try to add them to this top-level post. If someone is interested in implementing one or more of these, I can also put your name next to one of these. We can also discuss specifics of implementation in the comments below.

- [ ] Optional argument to append or overwrite when adding data selection with setters (Grace?)
- [ ] Get data as iterators (currently implemented in two different ways; need to settle on how to best do this) (Ian/Luigi)
- [ ] Create template waveforms over a selection
- [ ] Randomly sample waveforms within a selection
- [ ] Access to metadata and analysis parameters
- [ ] "Lazy" construction of the entry list for use with iterators and waveform browsers

gipert:
I also think that, since many new features will come, we should think about how to refactor the code to make it more readable. It's very hard at the moment for new people to understand the code and contribute (`data_loader.py` is huge!)

iguinn:
Another thing I'll add to the list is clearer error messages, and sanity checks as things are loading (with clear error messages)

erin717:
For accessing analysis paramters such as the filter parameters or calibration results, do we want that to be handled by the DataLoader or by pylegendmeta?

gipert:
I would argue that we should try to avoid having LEGEND-specific stuff in the DataLoader, such as pylegendmeta.

erin717:
I agree. I ask because if that's the case, we should remove or rethink the fifth bullet on the list as it broaches beyond the scope of what the data loader should be doing.

bossioel:
Hi all! I spent a couple of days trying to use the data loader to load the SiPM data, and I want to report some feedback (I hope this can be useful for future improvements).
- One problem in general (also for the germanium data) is that if you are trying to load a certain variable or cut on a certain variable that doesn’t exist, you don’t get any error message. I think it’s a problem, especially when you are applying some cuts, and you don’t know from the output if this is empty because you cut everything or because you are trying to look for something that doesn’t exist. (but I see there is already a similar point)
[...]

PR #493, opened 2023-05-15
Let ``DataLoader`` default setters overwrite (not append) + `lgdo.Struct.int_dtype`

gracesong312:
1. Adds the `key_dtype` to the Struct attributes so that we can use non-strings as keys in Struct. Converts to/from string when writing to/from disk. Also returns the output of `dl.load(merge_files=False)` as a Struct now that ints can be used as keys.
2. Changes the default behavior of DataLoader setter methods to overwrite existing parameters, but adds an `append` argument in case appending is the desired behavior. 

Addresses #492 first point


src/pygama/lgdo/lh5_store.py
```diff
@@ -344,7 +344,19 @@ def read_object(
                 # fields. If they all had shared indexing, they should be in a
                 # table... Maybe should emit a warning? Or allow them to be
                 # dicts keyed by field name?
-                obj_dict[field], _ = self.read_object(
+                if "keys_dtype" in h5f[name].attrs:
```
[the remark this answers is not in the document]
gracesong312: I agree that we don't need to support `bool`s or `float`s, but do you think we need some sort of flag that says whether or not to cast to an `int`? For example, if something was labeled as `'00001'` and you want to use that as a key instead of automatically casting it to the int `1`.

src/pygama/flow/data_loader.py
```diff
@@ -273,10 +275,10 @@ def set_files(self, query: str | list[str]) -> None:
         if not inds:
             log.warning("no files matching selection found")
 
-        if self.file_list is None:
-            self.file_list = inds
-        else:
+        if append and self.file_list is not None:
             self.file_list += inds
```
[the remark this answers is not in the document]
gracesong312: They were not, but I just added some changes that should fix this.

src/pygama/lgdo/lh5_store.py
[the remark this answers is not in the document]
gipert: Ok, let's add this extra attribute. Can you remove support for `int`s and `float`s?

gipert:
Could you add a test of the `append=False` functionality? Something that makes sure that a second call to a setter + loading does not fail (i.e. does not require a call to `.reset()`)/

src/pygama/lgdo/struct.py
```diff
@@ -44,17 +44,21 @@ def datatype_name(self) -> str:
...
-        return self.datatype_name() + "{" + ",".join(self.keys()) + "}"
+        return (
+            self.datatype_name() + "{" + ",".join([str(k) for k in self.keys()]) + "}"
+        )
 
     def update_datatype(self) -> None:
         self.attrs["datatype"] = self.form_datatype()
 
-    def add_field(self, name: str, obj: LGDO) -> None:
+    def add_field(self, name: str | int | float | bool, obj: LGDO) -> None:
```
gipert, line 54:
```suggestion
    def add_field(self, name: str | int, obj: LGDO) -> None:
```

src/pygama/lgdo/struct.py
```diff
@@ -44,17 +44,21 @@ def datatype_name(self) -> str:
...
         """Add a field to the table."""
         self[name] = obj
         self.update_datatype()
 
-    def remove_field(self, name: str, delete: bool = False) -> None:
+    def remove_field(
+        self, name: str | int | float | bool, delete: bool = False
```
gipert, line 60:
```suggestion
        self, name: str | int, delete: bool = False
```

src/pygama/flow/data_loader.py
```diff
@@ -299,12 +304,14 @@ def set_datastreams(self, ds: list | tuple | np.ndarray, word: str) -> None:
         word
             the type of identifier used in ds. Should be a key in the given
             channel map or a word defined in the configuration file.
+        append
+            if True, appends datastreams to the existing table_list instead of overwriting
```
gipert, line 308:
```suggestion
            if ``True``, appends datastreams to the existing `self.table_list`
            instead of overwriting.
```

src/pygama/flow/data_loader.py
```diff
@@ -244,6 +244,8 @@ def set_files(self, query: str | list[str]) -> None:
             supported by :meth:`pandas.DataFrame.query`. In addition, the
             ``all`` keyword is supported to select all files in the database.
             If list of strings, will be interpreted as key (cycle timestamp) list.
+        append
+            if True, appends files to the existing file_list instead of overwriting
```
gipert, line 248:
```suggestion
            if ``True``, appends files to the existing `self.file_list`
            instead of overwriting.
```

PR #494, opened 2023-05-16
Fix merging of FileDBs

erin717:
Merging of FileDBs introduced in #481 did not work as intended. Re-indexing only occurred for the first FileDB that had different columns. The re-indexing was also not properly implemented, leading to mismatches between {tier}_tables and {tier}_col_idxs. This PR fixes these issues.

Merge assumes the first FileDB in the list has less columns than the later FileDB's (such as from a calibration run). Sorts the list of FileDB's before merging to aid in this purpose.

### DIFF
--- a/src/pygama/flow/file_db.py
+++ b/src/pygama/flow/file_db.py
@@ -506,7 +506,7 @@ class FileDB:
 
         if not paths:
             raise FileNotFoundError(path)
-            
+
         # Need order of fileDBs to be consistent
         paths = sorted(paths)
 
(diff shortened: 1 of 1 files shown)


erin717:
Hey Luigi, 

I'm not sure I understand the problem you mentioned about making multiple substitutions and applying them in sequence. It's possible my solution fixes this, as I didn't encounter any errors with the dataframe merging, but I can't say for certain.

That being said, I propose we move forward with #495. I just tested it and it appears to fix the same issues I was aiming to fix, albeit in an admittedly cleaner way.

gipert:
Okay, I'm also adding unit tests to #495.

PR #500, opened 2023-05-26
Remove `pygama.{lgdo,raw,dsp}` subpackages

gipert:
The source code has moved to

* https://github.com/legend-exp/legend-pydataobj
* https://github.com/legend-exp/legend-daq2lh5
* https://github.com/legend-exp/dspeed

Corresponding tests and documentation have been removed as well.

importing `pygama.{lgdo,raw,dsp}` still works (I added wildcard imports from the new packages), but will emit a deprecation warning. The subpackages will be completely removed in the future.

This should not impact existing user programs, which should be however updated to import code from the new packages:
```
import pygama.lgdo -> import lgdo
import pygama.raw -> import daq2lh5
import pygama.dsp -> import dspeed
```
DSP configuration dictionaries should be updated to import processors from `dspeed` (and not `pygama.dsp`).

Also: closes #482.

PR #503, opened 2023-06-21
New moving linear fit processor for ppc a/e

ggmarshall:
Added a moving linear fit processor that can be used for the ppc a/e. Seems to give better results in testing.

### DIFF
--- a/src/pygama/dsp/processors/__init__.py
+++ b/src/pygama/dsp/processors/__init__.py
@@ -60,7 +60,7 @@ several advantages:
 """
 
 from .bl_subtract import bl_subtract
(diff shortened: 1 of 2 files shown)


gipert:
Since https://github.com/legend-exp/pygama/pull/500 is ready for merge I would propose to redirect this PR to https://github.com/legend-exp/dspeed

ggmarshall:
How long is that likely to take? I'd quite like to run a production with this in the next few days?

gipert:
I was just waiting for a final green light from @jasondet, we can merge it today

jasondet:
sorry I didn't see this last week. We just merged 500. Can this be moved to dspeed?

PR #514, opened 2023-09-29
A/E and calibration changes

ggmarshall:
Rewrite of a/e code, better calibration fitting with bias on tail to remove degeneracies.
Improved all results dictionaries to specify function fitted where it is found and parameters/errors as dictionaries.


gipert:
Is this ready to merge? You could mark the PR as "draft" and set it to "ready to be merged" when you want us to review and merge.

Also please install the pre-commit hooks locally with `pre-commit install`. In general, we want to avoid having these pre-commit commits in the Hit history. It would be nice if you could rewrite the current commit history by squashing the auto commits into the previous ones with `git rebase -i`.

ggmarshall:
I think this is ready to be merged now, there are more changes I want to make but I'll save them for other prs

ggmarshall:
I've also now changed the loading routines so we can use the tcm for pulser removal

gipert:
I have just realized some pre-commit checks don't get ran because the pargen directory is excluded in `.pre-commit-config.yaml`. It would be good to remove those exceptions. Merging this now.

PR #520, opened 2023-10-24
Bugfix for issue 516

MoritzNeuberger:
- Fixed the bug that `drop_idx` in `build_entry_list` was not filled if no cuts were set by removing the `continue` condition and making the default value of cut `cut = ""` which is changed if `self.cuts is not None` and `level in self.cuts.keys()`.
- Added `continue` condition if `self.table_list is not None` but not `level in self.table_list.keys()`. Otherwise the loop over `tables` would throw an error.


gipert:
This looks straightforward to me, I will merge it.

PR #525, opened 2023-11-07
Added routines for calibrating the LQ parameter

erin717:
Defines a general class that contains the different functions used for calibrating the Late Charge (LQ) parameter, in a similar manner to AoE_cal.py. A few plotting functions are redundant with those found in AoE_cal.py. We should consider generalizing some of these plotting functions in the future for other PSD parameters.

There are some improvements to be made in the future, but these routines are stable enough that it's worth implementing as things currently are. The routines have been tested individually on all detectors with no issues. However, I'll leave as a draft as there may be some small edits to make and it needs testing with the main data production.

### DIFF
--- a/src/pygama/pargen/lq_cal.py
+++ b/src/pygama/pargen/lq_cal.py
@@ -1,0 +1,938 @@
+from __future__ import annotations
+
+import json
+import logging
+import os
+import pathlib
+import re
+from datetime import datetime
+from typing import Callable
+
+import matplotlib as mpl
+mpl.use("agg")
+
+import matplotlib.cm as cmx
+import matplotlib.colors as mcolors
+import matplotlib.pyplot as plt
+import matplotlib.dates as mdates
+import numpy as np
+import pandas as pd
+from iminuit import Minuit, cost, util
+from scipy.stats import linregress
+from matplotlib.backends.backend_pdf import PdfPages
+from matplotlib.colors import LogNorm
+
+import lgdo.lh5_store as lh5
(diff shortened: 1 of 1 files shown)


src/pygama/pargen/lq_cal.py
```diff
@@ -0,0 +1,949 @@
...
+
+
+def plot_lq_cut_fit(lq_class, data, figsize=[12, 8], fontsize=12) -> plt.figure:
+    """Plots the final histogram of LQ values for events in the
+    DEP, and the fit results used for determining the cut
+    value"""
+
+    plt.rcParams["figure.figsize"] = figsize
+    plt.rcParams["font.size"] = fontsize
+    fig, (ax1, ax2) = plt.subplots(2, 1, height_ratios=(2, 1))
```
ggmarshall, line 714: needs mpl > 3.6, can either remove from here or update setup file

src/pygama/pargen/lq_cal.py
```diff
@@ -0,0 +1,949 @@
...
+        timestamp. Applies a time normalization based on the average LQ value
+        in the DEP across all timestamps.
+        """
+
+        log.info("Starting LQ time correction")
+        self.timecorr_df = pd.DataFrame(
+            columns=["timestamp", "mean", "mean_err", "res", "res_err"]
+        )
+        try:
+            if "timestamp" in df:
```
ggmarshall, line 284: timestamp -> run_timestamp in this function

ggmarshall (changes requested):

PR #527, opened 2023-11-08
Switching the dataloader build_entry_list from pandas to polars

ggmarshall:
seems to give a decent speedup in testing, definitely more optimisation possible as well with using Lazy dataframes. have left the cut query as pandas as this needs a parser to work with polars


gipert:
Jesus... what's this commit history?

ggmarshall:
I see no problem.....

ggmarshall:
It's because of the rebasing, will leave as this is just an exemplar

PR #531, opened 2023-11-21
Add support for bit compounds definitions in `build_hit()` config file

rosannadeckert:
added bit mask to build_hit and added test for bit mask

### DIFF
--- a/tests/hit/test_build_hit.py
+++ b/tests/hit/test_build_hit.py
@@ -96,8 +96,8 @@ def test_outputs_specification(dsp_test_file, tmptestdir):
     store = LH5Store()
     obj, _ = store.read_object("/geds/hit", outfile)
     assert list(obj.keys()) == ["calE", "AoE", "A_max"]
-    
-    
+
+
 def test_aggregation_outputs(dsp_test_file, tmptestdir):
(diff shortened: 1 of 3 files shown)


willquinn (approved):

tests/hit/test_build_hit.py
```diff
@@ -98,6 +98,33 @@ def test_outputs_specification(dsp_test_file, tmptestdir):
...
+        "aggr1",
+        "aggr2",
+    ]
+
+    df = store.load_dfs(outfile, ["aggr1", "aggr2"], "geds/hit/")
+
+    # aggr1 consists of 3 bits --> max number can be 7, aggr2 consists of 2 bits so max number can be 3
+    assert not (df["aggr1"] > 7).any()
+    assert not (df["aggr2"] > 3).any()
+
```
gipert, line 126: Can you add a test that checks whether each bit has the expected value? With this function, for example:
```py
def get_bit(x, n):
    return x & 1 << (n - 1) != 0
```
where `n` is the bit number (from least to most significant) and numbering starts from 1.

gipert: Lovely thank you!

rosannadeckert: I've added this test

PR #558, opened 2024-02-15
Added DPZ extractor to pargen

SamuelBorden:
The function `get_dpz_decay_constants` works to get the constants needed for the double pole zero filter in three steps:

1. A guess for the long and short time constants are obtained by taking linear fits to the logarithm of a superpulse
2. These guesses are used to seed a least squares fit 
3. The least squares values are then used in a dsp chain that optimizes the flatness of the decaying tails of waveforms

This is a first pass at the routine, so we should think maybe a bit more about where to put this in the dataflow. Right now this quickly interfaces with the single pole-zero dataflow script; to do that, the superpulse is created inside the function  `get_dpz_decay_constants`. But we might want to make `get_dpz_decay_constants` a bit more general and instead create the superpulse in the dataflow. 

There are other issues currently with the double pole zero filter, see https://github.com/legend-exp/dspeed/pull/52. So I am going to mark this as a draft PR


gipert:
@ggmarshall @SamuelBorden can this be merged to main or will mess up with the refactor?

SamuelBorden:
I need to change some imports from the energy cal routines, but then it should be good to go. I'll work on synergizing this with the pargen refactor

ggmarshall:
Can this be closed? Supplanted by #596 ?

PR #564, opened 2024-03-08
Pargen refactor

ggmarshall:
- [x] Energy cal plotting to dataflow
- [x] fix results output
- [x] Energy optimisation to new math
- [x] A/E to new math + add run by run sfs
- [x] LQ to new math
- [x] extract tau to class
- [x] Unit tests
- [ ] Notebooks


sagitta42:
Hi there, I'm already looking at some changes in this PR to note changes that I need to pay attention to in my code that uses classes and functions from pargen.

One thing that I noticed:

Before: `dep_acc` is a member of the `cal_aoe` class passed to init; in `cal_aoe.calibrate()` we call `cal_aoe.get_aoe_cut_fit()` which has an input variable `dep_acc` where we pass a hard coded calue of `0.9`; which is however not used, `self.dep_acc` is used inside that function.

Now: `dep_acc` is a member of the `CalAoE` class passed to init; but also an input variable in `CalAoE.calibrate()` which it then passes it to `CalAoE.get_aoe_cut_fit()`, but `self.dep_acc` is still used there anyway, not `dep_acc` passed by `calibrate()`.

It wouldn't break anything since it's always 0.9 as of now, but wanted to note this just in case, also from the point of view of `pars_hit_aoe.py` later.

sagitta42:
Another question:

I see that `ecal_th.py` is deleted entirely. I imagine there would be changes to the flow in `pars_hit_ecal.py` in a corresponding PR in `legend-dataflow`? Where did the `apply_cuts` and `calibrate_parameter` functionality get moved? (sorry, I can't seem to understand that from digging around in this PR) I am currently using both, following the procedure in `pars_hit_ecal.py`.

ggmarshall:
I'm still working my way up through the chain so haven't got to a/e, lq yet, the dataflow changes are coming but again I have been working on up to dsp and haven't finished higher tiers yet

sagitta42:
I see, thanks! I'll stay tuned then and prepare for changes :)

PR #568, opened 2024-03-21
Added numba defaults for math, fixed some bugs in extended pdfs

SamuelBorden:
This pull request adds numba defaults for the math distributions, like in `dspeed`, as requested by @ggmarshall for more user flexibility.

### DIFF
--- a/src/pygama/math/utils.py
+++ b/src/pygama/math/utils.py
@@ -3,7 +3,9 @@ pygama utility functions.
 """
 import logging
 import sys
-from typing import Optional, Union, Callable
+import os
+from typing import Optional, Union, Callable, Any, Iterator
+from collections.abc import MutableMapping
 
 import numpy as np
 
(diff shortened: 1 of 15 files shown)


src/pygama/math/utils.py
```diff
@@ -102,4 +104,76 @@ def print_fit_results(pars: np.ndarray, cov: np.ndarray, func: Optional[Callable
...
+    Override global options at runtime:
+
+    >>> from pygama.math.utils import numba_defaults
+    >>> # must set options before explicitly importing pygama.math.distributions!
+    >>> numba_defaults.cache = False
+    """
+
+    def __init__(self) -> None:
+        self.parallel: bool = getenv_bool("MATH_PARALLEL", default=True)
+        self.fastmath: bool = getenv_bool("MATH_FAST", default=True)
```
gipert, line 149: Why not moving this module to `pygama.utils`? We could use it for any Numba-decorated routine in pygama. Also, I would replace `MATH_` with `PYGAMA_` here.

src/pygama/utils.py
```diff
@@ -0,0 +1,76 @@
...
+    Override global options at runtime:
+
+    >>> from pygama.utils import numba_math_defaults
+    >>> # must set options before explicitly importing pygama.math.distributions!
+    >>> numba_math_defaults.cache = False
+    """
+
+    def __init__(self) -> None:
+        self.parallel: bool = getenv_bool("PYGAMA_PARALLEL", default=True)
+        self.fastmath: bool = getenv_bool("PYGAMA_FAST", default=True)
```
gipert, line 46:
```suggestion
        self.fastmath: bool = getenv_bool("PYGAMA_FASTMATH", default=True)
```

PR #576, opened 2024-04-26
Added math module back into pre-commit. Renamed classes following capitalization conventions.

SamuelBorden:
@gipert I've added the `math` module back into the pre-commit. I changed some variable names to all lowercase, and then changed all `math` distribution class names to uppercase. I fixed the imports related to this in `math` and `pargen` and fixed some docstrings that weren't compiling in `pargen`. Tests in `math` and `pargen` are still passing for me locally.


gipert:
lovely, thanks a lot sam!

PR #589, opened 2024-08-25
Updated pargen.utils.load_data to use LH5Iterator and field_mask to be more memory efficient

iguinn:
`load_data` was using too much memory (at least in P08 which is a larger dataset), and crashing my attempts to process it on NERSC. Switch it to use the fieldmask when reading from the input files and the LH5Iterator to limit the number of entries in memory at once. Also improved commenting/docstring.

Note this should not be merged until this is also merged: https://github.com/legend-exp/legend-pydataobj/pull/100

### DIFF
--- a/src/pygama/pargen/utils.py
+++ b/src/pygama/pargen/utils.py
@@ -81,17 +81,18 @@ def load_data(
                 threshold,
                 return_selection_mask,
             )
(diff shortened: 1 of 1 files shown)


ggmarshall:
Not on my end but I can have a look later this week

PR #591, opened 2024-10-09
A/E survival fractions joint fitter and ecal fix

ggmarshall:
Added new joint fitter for survival fractions which is in own file. 
Improved A/E fitting so now more stable, enabling run to run corrections.
Added two new modes to the time correction: average_consecutive and interpolate_consecutive
Improved docstrings
Small tweaks to ecal: upper bound on sigma and some minor sigma guess changes to remove a few edge cases.

### DIFF
--- a/src/pygama/pargen/AoE_cal.py
+++ b/src/pygama/pargen/AoE_cal.py
@@ -719,8 +719,10 @@ def get_survival_fraction(
 
     pc_n = ct_n + surv_n
 
-    sf = (surv_n / pc_n) * 100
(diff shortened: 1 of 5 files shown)


erin717:
We should consider moving the survival fraction code out of the A/E calibration script. It has enough general purpose utility beyond A/E to warrant being it's own thing.

ggmarshall:
Yep I agree will move into our script

ggmarshall:
@iguinn This changes should make A/E run by run corrections viable moving forward as well as cleaning up a couple of bugs and hopefully a more stable survival fraction calculation

PR #596, opened 2024-12-16
Add double-pole-zero routines

erin717:
Adds the double-pole-zero routines written by Sam Borden (see #558 for more details). Updated the routines to work with the pargen refactor. Removed the old dpz grid search routines. Added unit tests.

**Note:** The dpz routine ends with an iMinuit minimization for the flatness of a dpz-corrected waveform. Currently, the double pole zero processor in dspeed throws an exception if the output is nans, During the minimization, some function calls have nan outputs. This issue will need to be addressed (either by removing the exception or forgoing the minimization step)

### DIFF
--- a/src/pygama/pargen/extract_tau.py
+++ b/src/pygama/pargen/extract_tau.py
@@ -4,7 +4,6 @@ This module is for extracting the pole zero constants from the decay tail
 
 from __future__ import annotations
 
-import logging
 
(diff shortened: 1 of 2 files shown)


ggmarshall:
Can we just catch the error in the optimiser in dspped?

ggmarshall:
Superseded by #603

PR #597, opened 2025-02-05
Changed `evt/spms` to allow for event building with zero-suppression data

rosannadeckert:
- Fixed bug in `hit/build_hit.py` 
- Changed functions in `evt/spms.py` to account for data in zero-suppression mode, i.e. when each event has a different number of channels that triggered

### DIFF
--- a/src/pygama/evt/modules/spms.py
+++ b/src/pygama/evt/modules/spms.py
@@ -88,10 +88,29 @@ def gather_pulse_data(
         lgdo_obj = lh5.read(
             f"/{channel}/{tierinfo.group}/{column}", tierinfo.file, idx=idx
         )
-        data = lgdo_obj.view_as(library="ak")
+        data = lgdo_obj.view_as(library="np")
 
         # remove nans (this happens when SiPM data is stored as ArrayOfEqualSizedArrays)
         data = ak.drop_none(ak.nan_to_none(data))
+
+        # number of channels per event
+        evt_length = np.diff(np.insert(tcm.cumulative_length, 0, 0))
+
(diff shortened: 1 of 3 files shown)


src/pygama/evt/modules/spms.py
```diff
@@ -88,11 +88,30 @@ def gather_pulse_data(
...
+        # construct global event ID's
+        glob_ids = np.repeat(
+            np.arange(0, tcm.cumulative_length.shape[0], 1), evt_length
+        )
+        glob_ids_ch = glob_ids[
+            tcm.id == table_id
+        ]  # global ID's where channel had a trigger
+
+        # count number of hits per channel for global events with trigger in channel, else 0
+        glob_ids_cts = np.zeros(tcm.cumulative_length.shape[0], dtype="int8")
```
gipert, line 108: isn't this `int8` dangerous?

rosannadeckert: what should i put?

gipert: I would just put `int` (the Python one, which is analogous to `numpy.int32`)

src/pygama/evt/modules/spms.py
```diff
@@ -121,7 +140,7 @@ def gather_pulse_data(
             pulse_mask = pulse_mask.view_as("ak")
 
         # apply the mask
-        data = data[pulse_mask]
+        data = data[ak.values_astype(pulse_mask, bool)]
```
gipert, line 143: Note to myself: check what's going on here

PR #599, opened 2025-03-11
tcm handle multiple cols, change to vov

ggmarshall:
- [x] update pargen
- [x] update evt
- [x] update testdata
- [x] moar tests
- [x] multiple channel entries in single tcm
- [x] multiple channel entries in single event
- [x] rename tcm cols

needs: https://github.com/legend-exp/legend-testdata/pull/26

SiPM changes need testing, maybe add unit tests?


ggmarshall:
@tdixon97 I've updated evt to be similar to reboost if you can take a look

ggmarshall:
One thing not considered is multiple entries for the same channel in an event

PR #609, opened 2025-05-06
Small change in dplms routine

valerioda:

### DIFF
--- a/src/pygama/pargen/dplms_ge_dict.py
+++ b/src/pygama/pargen/dplms_ge_dict.py
@@ -14,7 +14,6 @@ from lgdo import Table
 from scipy.signal import convolve, convolve2d
 from scipy.stats import chi2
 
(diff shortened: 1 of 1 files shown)


src/pygama/pargen/dplms_ge_dict.py
```diff
@@ -108,7 +106,7 @@ def dplms_ge_dict(
         "peak": peaks_kev[-1],
         "kev_width": kev_widths[-1],
         "parameter": ene_par,
-        "func": gauss_on_step,
+        "func": eval(dplms_dict["fom_func"]),
```
ggmarshall, line 109: maybe better to add func as an arg for the function? can handle this eval in the dataflow which keeps this code clean?

valerioda: Yes, I think this is a good idea

PR #613, opened 2025-06-12
Remove legend-daq2lh5 from the dependencies and old stuff from the CLI

gipert:
This dependency is unnecessary. Also remove remnants of pygama cli interface to dspeed and legend-daq2lh5.


gipert:
😮‍💨

PR #634, opened 2025-11-25
add pulser scan analysis module

fraborra:
Provides pulser scan analysis methods.
Exposed as both standalone functions and via a wrapper class to improve usability and flexibility.


valerioda:
What is the status of this PR? Should be merged?

PR #680, opened 2026-05-11
add Bayesian-blocks adaptive rebinning to pygama.math

gipert:
## Summary

- Adds `pygama.math.rebin` with two entry points backed by `astropy.stats.bayesian_blocks`:
  - `hist_bblocks(data, *, prebin_width=None, prebin_low=None, prebin_high=None, p0=0.05)` — adaptive histogram of an unbinned data array (numpy or awkward). `prebin_width` activates a pre-binning step (fine `Regular` axis → `rebin_bblocks`) that brings the O(N²) `events`-fitness cost down to O(N_bins²); `prebin_low`/`prebin_high` bound the pre-binned range.
  - `rebin_bblocks(h, *, p0=0.05)` — rebin a finely-binned 1D `hist.Hist` using `fitness='measures'`; snaps block edges to the input grid and aggregates counts/variances via `np.add.reduceat`, so output edges are an exact subset of the input edges and total counts/variances are conserved.
- Adds `astropy` and `awkward` as runtime dependencies.

## Test plan

- [x] `pytest tests/math/test_rebin.py -v` — 9 tests pass
  - `hist_bblocks` on numpy / awkward inputs and on two-population data
  - pre-binning path: count preservation, gap detection, range restriction, error when `prebin_low`/`prebin_high` is set without `prebin_width`
  - `rebin_bblocks` total counts / variances preserved
  - `rebin_bblocks` output edges are an exact subset of the input edges (both regular and irregular input grids)
  - flat input collapses to ≤ 3 blocks
- [x] `pre-commit run --files src/pygama/math/rebin.py tests/math/test_rebin.py pyproject.toml` — clean
- [x] `cd docs && make` — both functions appear in `build/html/api/pygama.math.html` with no Sphinx warnings about the new module

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/src/pygama/math/rebin.py
+++ b/src/pygama/math/rebin.py
@@ -1,0 +1,180 @@
+"""Bayesian-blocks adaptive rebinning for 1D histograms.
+
+Provides routines for adaptive binning based on the Bayesian-blocks
+algorithm (Scargle et al. 2013), backed by
+:func:`astropy.stats.bayesian_blocks`.
+
+Two entry points are provided:
+
+- :func:`hist_bblocks` histograms an unbinned data array
+  (:class:`numpy.ndarray` or :class:`awkward.Array`) using
+  Bayesian-blocks edges.
+- :func:`rebin_bblocks` rebins a finely-binned :class:`hist.Hist`
+  using the ``measures`` fitness.
+
+Both return a ``Variable``-axis ``Weight``-storage :class:`hist.Hist`.
+"""
+
+from __future__ import annotations
+
+import awkward as ak
+import hist as bh
+import numpy as np
+from astropy.stats import bayesian_blocks
+
+
+def hist_bblocks(
+    data: np.ndarray | ak.Array,
+    *,
+    prebin_width: float | None = None,
+    prebin_low: float | None = None,
+    prebin_high: float | None = None,
+    p0: float = 0.05,
+) -> bh.Hist:
+    """Histogram an unbinned data array using Bayesian-blocks edges.
+
+    By default uses :func:`astropy.stats.bayesian_blocks` with
+    ``fitness='events'`` directly on ``data``. The ``events`` fitness
+    is O(N²) in the number of points, which becomes prohibitive
+    for large samples; passing ``prebin_width`` first bins the data into
+    a uniform fine histogram of that width and then runs
+    :func:`rebin_bblocks` (``measures`` fitness) on it, which is
+    O(Nₙ²) in the number of fine bins.
+
+    Awkward inputs are flattened across all axes before binning.
+
+    Parameters
+    ----------
+    data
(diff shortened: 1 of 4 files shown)


gipert:
@sofia-calgaro @tdixon97 i think i am satisfied with this, if you want to have a look

gipert:
image
image

tdixon97:
I like it, its very useful. Can you add a little bit of documentation? Even if its just an minimal example?

tdixon97:
@gipert is it ready? Id like to try with the ssc / rdc data

gipert:
me too that's why i kept it open, there is maybe stuff to tweak still

tdixon97:
can you just add some basic description of what functions to use etc? I can use this branch.

I think the summary above is almist enough

gipert:
aren't the docstrings good enough?

tdixon97:
kinda .. but I do not know what the functions are / what they are called.

src/pygama/math/rebin.py
```diff
@@ -0,0 +1,626 @@
...
+    for li, ri in regions:
+        if merged and li <= merged[-1][1]:
+            merged[-1][1] = max(merged[-1][1], ri)
+        else:
+            merged.append([li, ri])
+
+    return np.array([[fine_edges[li], fine_edges[ri + 1]] for li, ri in merged])
+
+
+def _collapse_peaks(
```
tdixon97, line 218: I think this could be a public function

src/pygama/math/rebin.py
```diff
@@ -0,0 +1,626 @@
...
+
+    keep = np.ones(len(fine_edges), dtype=bool)
+    for lo_p, hi_p in regions:
+        li = int(np.searchsorted(fine_edges, lo_p))
+        ri = int(np.searchsorted(fine_edges, hi_p))
+        keep[li + 1 : ri] = False
+    # always route through bh.rebin so the output axis is uniformly Variable,
+    # even when no consolidation happens (a plain copy would preserve a
+    # Regular axis input, violating the module contract).
+    return h[:: bh.rebin(edges=fine_edges[keep].tolist())]
```
tdixon97, line 242: Might be nice to have a convience function to rebin another histogram / data with the same binning

src/pygama/math/rebin.py
gipert: The reason why I set it private is that this function only works if the input is a histogram already binned with bayesian blocks. We can make it public if you think it's better but then we need to improve the function name

src/pygama/math/rebin.py
gipert: I haven't tested it but should work with sth like

```py
h[::bh.rebin(edges=other_hist.edges)]
```

PR #686, opened 2026-06-01
Reduce memory usage of noise optimisation

ggmarshall:
- [x] Remove fft field when not needed — superseded: the grid-search loop now requests **only** the configured `ene_str` outputs, which drops the fft field implicitly along with every other unread output buffer (most notably `wf_presum`, a waveform-length output of ~160 MB at the production `n_events=10000`, previously reallocated on every grid point).
- [x] Make the fft plot process in batches to reduce memory — the lgdo requirement (`Table`/`WaveformTable` slicing, legend-exp/legend-pydataobj#212) landed in legend-pydataobj v2.0.0, inside the existing `>=2` pin, so no pin change was needed.

Also in this PR: the FFT plot's frequency axis previously used `wf_presum.dt`, but the PSD is computed on `wf_blsub_full`, which derives from the full `waveform` — the axis was off by the presum factor. It now uses `tb_data['waveform'].dt`.

### DIFF
--- a/src/pygama/pargen/noise_optimization.py
+++ b/src/pygama/pargen/noise_optimization.py
@@ -81,6 +81,7 @@ def noise_optimization(
         plot_dict["nopt"] = {"fft": {"frequency": freq, "psd": psd, "fig": fig}}
         plt.close()
 
+    dsp_proc_chain = {**dsp_proc_chain, "outputs": dsp_proc_chain["outputs"].copy()}
     if opt_dict.get("fft_field", "wf_psd") in dsp_proc_chain["outputs"]:
         dsp_proc_chain["outputs"].remove(opt_dict.get("fft_field", "wf_psd"))
 
--- a/src/pygama/pargen/noise_optimization.py
+++ b/src/pygama/pargen/noise_optimization.py
@@ -112,9 +112,12 @@ def noise_optimization(
         plot_dict["nopt"] = {"fft": {"frequency": freq, "psd": psd, "fig": fig}}
         plt.close()
 
-    dsp_proc_chain = {**dsp_proc_chain, "outputs": dsp_proc_chain["outputs"].copy()}
-    if opt_dict.get("fft_field", "wf_psd") in dsp_proc_chain["outputs"]:
-        dsp_proc_chain["outputs"].remove(opt_dict.get("fft_field", "wf_psd"))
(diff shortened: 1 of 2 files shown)


ggmarshall:
Implemented the batching in 50912c38, closing the last checklist item.

`_batched_mean_psd` slices the table into `opt_dict['fft_batch_size']` chunks (default 1000) and requests only the fft field from the chain, so the energy filters are skipped for the plot run and the mean accumulates in float64. Verified on a real p18 fft channel (1438 baselines, 32768-sample waveforms, the production minimal chain):

| | old (full-table run) | batched (256/batch) |
|---|---|---|
| mean PSD | — | identical to float32 roundoff (max rel diff 2.3e-6) |
| time | 47.2 s | 3.8 s |
| peak RSS | 1.04 GB | 0.70 GB |

The old path's footprint grows linearly with `n_events` (production config uses 10 000, ~2 GB for `wf_blsub_full`+`wf_psd` alone); the batched one is flat in it. Six unit tests cover batching equivalence, the display path end-to-end (including that the caller's proc chain stays unmutated), the empty-input and non-positive-batch-size errors, and a custom `fft_field` name.

ggmarshall:
Follow-up in 5c8d03fb: the grid-search loop now requests only the `ene_str` fields from the proc chain instead of the config's full outputs list minus the fft field. The dependency map makes this safe — the loop reads nothing but `dsp_data[ene_str]`, and the energy filters run entirely on the presummed waveform — and it removes the `wf_presum` output buffer (~160 MB at `n_events=10000`) that was reallocated on each of the ~28 grid points.

Measured on the same p18 fft channel as before (n=1438): grid-iteration peak RSS 0.67 → 0.58 GB, with `trapEmax`/`cuspEmax`/`zacEmax` **bit-identical** to the full-outputs run. A side effect worth noting: configs no longer need to list the `ene_str` fields in `outputs` for the optimisation to run.

One further gain was identified but belongs in legend-dataflow-scripts, not here: the raw table read in `nopt.py` loads all columns, while the grid loop needs only `waveform_presummed` (+`presum_rate`/`baseline`) — the full `waveform` (64 KB/event, ~75% of the table) is used only by the FFT plot and `waveform_windowed` by nothing. A `field_mask` there would cut the dominant term from ~0.88 GB to ~0.2 GB at production scale for `display=0` runs.

PR #713, opened 2026-08-29
Adding geds cross talk routine to pygama.pargen

hungwei59079:
Added geds xtalk production routine using amplitude method to pygama.pargen. 

The amplitude method uses the following formula to compute the xtalk values between geds:

$$
(R^{\mathrm{norm}}_{t \to r})_i
= \frac{(E_{\mathrm{resp}})_i - \langle E_{\mathrm{bl}} \rangle}
{(E_{\mathrm{trig}})_i}
$$

The energies are expressed with uncalibrated energies, currently default to `trapTmax` for trigger detector, and `trapTmax` (`trapTmin`) for the response detector in positive (negative) xtalk computation. $\langle E_{\mathrm{bl}} \rangle$ represents the average baseline of the response detector. Such value is evaluated for every event that meets the following criteria: (i) The trigger detector is fired with high enough energy (a default threshold >1500keV is used, and I also included a qc flag is_highly_positive_polarity_candidate) (ii) The response detector is not fired (a default threshold <100 keV is used). After that, we fill the values into a histogram and fit with Gaussian to get the (average) xtalk value. 

The newly added script is designed to be processed with two rules: The first step assigns N jobs, where N equals to the number of detectors. Each job runs the function `prepare_detector` once, which computes the baseline of each detector and does a selection to compute which events to keep when this detector is treated as a trigger or response detector. An index array and a boolean mask is computed and returned following the selection. 

Next, the second step submits one single job that runs `xtalk_element` recursively for each pair of detectors to fit the xtalk values and assemble the xtalk values with `build_xtalk_matrix`. The matrix is then returned in the form of a lh5 file, and the xtalk matrix could be plotted with the utility `plot_xtalk_matrix`.

The dataflow script that implements these two rules are not in legend-dataflow yet. I haven't opened a PR to legend-dataflow either, so the script currently lives in my experimental branch: https://github.com/hungwei59079/pygama-fork/blob/experimental/dataflow_draft/xtc.py


valerioda:
Thanks a lot for this PR, following what has been done in the other pargen scripts, the configuration parameters and the selection of the data needed for the xtalk extraction can be done externally, to keep the pygama script more general

